PR #340, opened 2014-12-27
Recurring PRO deals

vevix:
:point_up: 

Will allow users to redeem a new code 1 month after their previous claim on deals that are recurring.

### DIFF
--- a/app/controllers/partner_deals_controller.rb
+++ b/app/controllers/partner_deals_controller.rb
@@ -16,8 +16,7 @@ class PartnerDealsController < ApplicationController
     # has this user redeemed a code?
     code = deal.codes.where(user: current_user).last
     if code.nil? || (deal.recurring? && Time.now > deal.recurring.seconds.since(code.claimed_at))
-      code = deal.codes.unclaimed.first
-      code.update_attributes!(user: current_user, claimed_at: Time.now)
+      deal.codes.unclaimed.limit(1).update_all(user_id: current_user.id, claimed_at: Time.now)
     end
 
     render json: deal, serializer: PartnerDealSerializer
--- a/app/controllers/partner_deals_controller.rb
+++ b/app/controllers/partner_deals_controller.rb
@@ -15,7 +15,7 @@ class PartnerDealsController < ApplicationController
     deal = PartnerDeal.find(params[:id])
     # has this user redeemed a code?
     code = deal.codes.where(user: current_user).last
-    if code.nil? || (deal.recurring? && Time.now > 1.month.since(code.claimed_at))
+    if code.nil? || (deal.recurring? && Time.now > deal.recurring.seconds.since(code.claimed_at))
       code = deal.codes.unclaimed.first
       code.update_attributes!(user: current_user, claimed_at: Time.now)
     end
--- a/app/controllers/partner_deals_controller.rb
+++ b/app/controllers/partner_deals_controller.rb
@@ -15,16 +15,12 @@ class PartnerDealsController < ApplicationController
     deal = PartnerDeal.find(params[:id])
     # has this user redeemed a code?
     code = deal.codes.where(user: current_user).last
-    if code.nil?
-      # todo: deal with potential race condition
+    if code.nil? || (deal.recurring? && Time.now > 1.month.since(code.claimed_at))
       code = deal.codes.unclaimed.first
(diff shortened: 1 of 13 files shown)


vikhyat:
@vevix what is the status of this?

db/structure.sql
```diff
@@ -1088,8 +1088,8 @@ CREATE TABLE partner_deals (
     deal_url text NOT NULL,
     deal_description text NOT NULL,
     redemption_info text NOT NULL,
-    recurring boolean DEFAULT false NOT NULL,
-    active boolean DEFAULT true NOT NULL
+    active boolean DEFAULT true NOT NULL,
+    recurring integer DEFAULT 0
```
NuckChorris: Probably need to rerun `bundle exec annotate` since you changed the db schema

vevix:
Should be good after the rebase, :balloon:

app/serializers/partner_deal_serializer.rb
```diff
@@ -1,12 +1,23 @@
...
+  end
+
+  def claimed_at
+    code_object && code_object.claimed_at
+  end
+
+  private
+
+  def code_object
+    scope && object.codes.where(user: scope).last
```
NuckChorris: This seems like a bad idea to use queries directly in a serializer.  Is there some other way this could be done?

vevix: Not sure it can be done any other way, but its not like we don't query in other serializers ¯\\\_(ツ)_/¯

app/frontend/app/models/partner-deal.js
```diff
@@ -7,7 +9,22 @@ export default DS.Model.extend({
...
+  recurring: DS.attr('number'),
   hasCodes: DS.attr('boolean'),
-  code: DS.attr('string')
+  code: DS.attr('string'),
+  claimedAt: DS.attr('date'),
+
+  isRecurring: Ember.computed.gt('recurring', 0),
+
+  canRedeemAgain: function() {
+    var date = new Date(this.get('claimedAt'));
```
NuckChorris: Ember won't automatically give a `Date` since it's a `'date'` field?

vevix: It's been awhile but looking at it, using `setSeconds` may have caused it to edit the property directly.

app/controllers/partner_deals_controller.rb
```diff
@@ -15,16 +15,12 @@ def update
     deal = PartnerDeal.find(params[:id])
     # has this user redeemed a code?
     code = deal.codes.where(user: current_user).last
-    if code.nil?
-      # todo: deal with potential race condition
```
NuckChorris: I'm pretty sure moving this to the `before_save` won't fix the race condition of two users claiming at the same time, since Rails doesn't know the state has changed in the database.

Perhaps an alternative solution would be to (ab)use `update_all`? `deals.codes.unclaimed.first.update_all(user: current_user, claimed_at: Time.now)` (though I'm not sure how the lack of AR typecasting on `update_all` might affect that with the use of Time and User)

app/frontend/app/models/partner-deal.js
NuckChorris: Ah, makes sense.  Sometimes I forget JS doesn't have a handy `dup` method :stuck_out_tongue:

app/controllers/partner_deals_controller.rb
vevix: :+1: I'll push this change tonight.

NuckChorris: Once that's in there, consider this LGTM'd.  Looks great, even if I don't think we have a use... yet!

vevix:
@NuckChorris Pushed the code update, I had to change the test to 2 months because February was bugging it out being only 28 days.

NuckChorris:
huh, does that mean we'll get weird results in short months?  I thought `1.month` always returned `30.days`

vevix:
```
irb(main):002:0> 30.days.since(1.month.ago)
=> Wed, 11 Mar 2015 19:32:54 UTC +00:00
```

Well it just means that it will allow recurring every 30 days instead of every "month". Working as intended IMO :cat:

NuckChorris:
Resolve the merge conflict and LGTM?

PR #349, opened 2015-01-04
Add title_language_preference parameter to APIv1 search method

erengy:
All anime and library methods support this parameter, except the search method. This results in confusing behavior for clients, where the API returns main titles differently for each method.

Ideally, the API would provide all available titles (as in APIv2) and let the client decide which one to use as the main title. But since we don't want to break APIv1's backwards compatibility, all we can do is to make it more consistent in this case.


vikhyat:
Thanks for this!

PR #351, opened 2015-01-05
Update index.hbs

skcihmit:
Add {{#unless notYetAired}} to Right Side in order to remove ratings on anime pages for not yet aired shows.

### DIFF
--- a/app/frontend/app/templates/anime/index.hbs
+++ b/app/frontend/app/templates/anime/index.hbs
@@ -233,6 +233,7 @@
   <!-- Left side End -->
 
   <!-- Right Side Start -->
+  {{#unless notYetAired}}
   <div class="community-column">
     {{#if roundedBayesianRating}}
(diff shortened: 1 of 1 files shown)


app/frontend/app/templates/anime/index.hbs
```diff
@@ -233,6 +233,7 @@
   <!-- Left side End -->
 
   <!-- Right Side Start -->
+  {{#unless notYetAired}}
   <div class="community-column">
```
Wopian: `div` and the following elements should probably be indented too to follow the layout already used:

```hbs
{{#if roundedBayesianRating}}
       <div class="hb-score">
              ...
       </div>
{{/if}}
```

skcihmit: Changed. Thanks.

vikhyat:
Thanks for this! :heart:

PR #353, opened 2015-01-06
Fix text color for quick update

fractalemagic:
The anchor element overrides the text color declared by `series-title`. This is a regression that happened a *long* time ago that always annoyed me.

Change Preview


vikhyat:
Not sure if this was intentional, @JoshFabian?

JoshFabian:
Wasn't intentional, it's annoyed me for quite some time but not enough to do something about it. Thanks! :)

vikhyat:
awesome

Issue #345, opened 2014-12-30
User about's panel its blank when not value. 


nivla:
I think that the panel "About User" its must have a default value like "Nothing yet" rather than leave it blank the panel.
![screenshot-hummingbird me 2014-12-30 09-20-34](https://cloud.githubusercontent.com/assets/2448098/5578828/b38aa244-9005-11e4-9c9a-e2d584326f34.png)

cybrox:
I don't think just filling in a default value here would make sense.
Guess the best solution would be still displaying it for the user itself with an info message to add some content and hide it for profile visitors.

cybrox:
Fixed with #358

PR #358, opened 2015-01-09
About section fix

cybrox:
This adds a default message for the **About user** section to the user profiles.
The message is different, depending on if you're viewing your own profile or not.
*Will fix* #345


nivla:
Nice!!!

vevix:
:+1: Looks good!

PR #360, opened 2015-01-11
Switch rails console to Pry (Fixes #323)

### LINKED ISSUE #323: Change Rails Console to Pry
Rails Console sucks, it'd be nice to have [Pry](http://pryrepl.org) instead.

NuckChorris:
Title says it all.  Turns out all you gotta do is require the `pry-rails` gem.


cybrox:
lgtm

Issue #357, opened 2015-01-09
Don't show user actions to guests on anime/manga pages


Wopian:
![](https://i.imgur.com/GppH0EE.png?1)

At the moment it's visible, but changing it just throws a connection error alert.

*Will send PR once done with school.*

Wopian:
Fixed in #361

cybrox:
Speaking of disabling user actions, this might be something that could be implemented to the follow buttons as well.
I guess genereally hiding all the login required UI elements would make more sense than just throwing the user a `you need to be logged in` message.
Thoughts?

vikhyat:
I think we should be showing a sign in modal instead of hiding half of our UI.

PR #361, opened 2015-01-11
Don't show user actions to guests on anime pages

Wopian:
Star ratings are hidden for guests, but the dropdown button is still visible and just throws a connection error alert when a status is selected.

#357


NuckChorris:
:+1:

cybrox:
lgtm

PR #364, opened 2015-01-13
Favorite refactoring (wip)

cybrox:
### Stuff to do
- [x] Refactor ember favorite handling (polymorphic)
- [x] Refactor user controller for anime and manga favorites
- [x] Add global component for favorite box of any media type
- [x] Implement drag and drop sort
- [x] Delete favorite records via ember data
- [x] Bulk-update `fav_rank` property of favorites
- [x] Fuzzy search anime and manga input to add media from fav box
- [x] Build new polymorphic relation in ember for new favorites (*undocumented ember shizzle*)
- [ ] Add endpoint to effectively add media to your favorites
 - Anime and Manga is using id as identifier for their polymorphic relation on rails side, while the ember application is using slugs as ids. This can either be solved by a workaround or maybe a direct way with Rails tricks. (cc. @NuckChorris)
- [ ] Make it look pretty
 - The whole UI needs **lots** of CSS love (cc. @JoshFabian) 

=====
- [ ] Add favorite button to manga pages
 - This is optional, since these buttons might get removed
- [ ] Add tab switch or new box for manga favorites
 - This should be done pretty quickly. We need a second edit indicator then though. Maybe this PR can be merged without this for now to test the new favorite box for anime and add all the manga stuff a bit later.


cybrox:
Moved to a seperate branch on the main repo for easy group work

PR #366, opened 2015-01-14
Users to follow on onboarding

### LINKED ISSUE #331: Onboarding "Who To Follow" should be hooked up
The "Who To Follow" is literally just showing whoever happens to be in the ED store at the time... we should probably make it actually display something useful, maybe use the same old featured-user flag that ran the "Who To Follow" sidebar?

cybrox:
Using the old user `to_follow` parameter to only show these users on the `onboarding.finish` page.
This implements:
- Kotodama: add and remove user to follow.
- Onboarding->Finish: displya only users with the to_follow flag.

This will fix #331

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -234,7 +234,10 @@ class UsersController < ApplicationController
       user.waifu_or_husbando = changes[:waifu_or_husbando]
       user.bio = changes[:mini_bio] || ""
       user.waifu_char_id = changes[:waifu_char_id]
-      user.to_follow = changes[:to_follow]
+
+      if current_user.admin?
+        user.to_follow = changes[:to_follow]
+      end
 
       if Rails.env.production? and changes[:cover_image_url] =~ /^data:image/
         user.cover_image = changes[:cover_image_url]
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -17,7 +17,7 @@ class UsersController < ApplicationController
       render json: users, meta: {cursor: 1 + (params[:page] || 1).to_i}
 
     elsif params[:to_follow]
-      render json: User.where(:to_follow => true), each_serializer: UserSerializer
+      render json: User.where(to_follow: true), each_serializer: UserSerializer
 
     else
       ### OLD CODE PATH BELOW. Used only by the recommendations page.
@@ -234,10 +234,6 @@ class UsersController < ApplicationController
       user.waifu_or_husbando = changes[:waifu_or_husbando]
       user.bio = changes[:mini_bio] || ""
       user.waifu_char_id = changes[:waifu_char_id]
-
-      if current_user.admin?
-        user.to_follow = changes[:to_follow]
-      end
 
       if Rails.env.production? and changes[:cover_image_url] =~ /^data:image/
         user.cover_image = changes[:cover_image_url]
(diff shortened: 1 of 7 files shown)


vikhyat:
We will need to backfill the current set of users?

cybrox:
Yes, we only have ~8 users in there though, so it shouldn't be a problem

app/frontend/app/controllers/kotodama.js
```diff
@@ -117,6 +125,30 @@ export default Ember.ObjectController.extend({
...
     toggleNonMal: function() {
       this.toggleProperty('showNonMal');
+    },
+
+    addUserToFollow: function(){
+      var username = this.get('userToFollow'),
+          self = this;
+
+      this.store.find('user', username).then(function(user){
+        if(user.length === 0){ return; }
```
vikhyat: :space_invader: before the `{`, here and in other places.

app/controllers/users_controller.rb
```diff
@@ -16,6 +16,9 @@ def index
 
       render json: users, meta: {cursor: 1 + (params[:page] || 1).to_i}
 
+    elsif params[:to_follow]
+      render json: User.where(:to_follow => true), each_serializer: UserSerializer
```
vikhyat: New style hash syntax would be better, `to_follow: true`.

app/frontend/app/controllers/kotodama.js
```diff
@@ -8,6 +8,8 @@ export default Ember.ObjectController.extend({
   blotterMesgFix: "",
   blotterMesg: "",
   blotterLink: "",
+  userToFollow: "",
+  userToFollowSubmit: false,
```
vikhyat: Maybe this should be `userToFollowSubmitted` for more clarity, since it is a state variable?

app/frontend/app/models/user.js
```diff
@@ -18,6 +18,7 @@ export default DS.Model.extend({
   titleLanguagePreference: DS.attr('string'),
   isAdmin: DS.attr('boolean'),
   isPro: DS.attr('boolean'),
+  toFollow: DS.attr('boolean'),
```
vikhyat: Do we need this here? This field would be added for _all_ users.

app/serializers/user_serializer.rb
```diff
@@ -2,7 +2,8 @@ class UserSerializer < ActiveModel::Serializer
   attributes :id, :cover_image_url, :avatar_template, :rating_type, :mini_bio,
     :about, :is_followed, :title_language_preference, :location, :website,
     :waifu, :waifu_or_husbando, :waifu_slug, :waifu_char_id, :last_sign_in_at,
-    :current_sign_in_at, :is_admin, :following_count, :follower_count, :is_pro
+    :current_sign_in_at, :is_admin, :following_count, :follower_count, :is_pro,
+    :to_follow
```
vikhyat: Again, do we really need to include `to_follow`? If possible we should avoid it.

app/frontend/app/templates/kotodama.hbs
```diff
@@ -90,6 +90,34 @@
...
+            </div>
+          </div>
+          <table class="table table-striped">
+            <tbody>
+              {{#each user in usersToFollow}}
+                <tr><td>
+                  <span class="pull-right delete">
+                    <i class="fa fa-minus-square-o" {{action 'removeUserToFollow' user}}></i>
+                  </span>
+                  {{#link-to 'user' user.id}}{{user.id}}{{/link-to}}
```
vikhyat: Use {{user.username}} inside the link text. Also does `{{#link-to 'user' user}}` not work?

vikhyat:
Awesome, let's :ship: this.

PR #371, opened 2015-01-15
Update ember-cli from 0.1.3 to 0.1.7

cybrox:
According to http://www.ember-cli.com/#upgrading


NuckChorris:
I don't see much which could blow things up, outside of the beta bump on `ember-data` and the patch bumps on various Rails gems.  :+1: from me.

vikhyat:
Maybe also upgrade the ember-cli-rails gem and ember adding while we're doing this?

cybrox:
@vikhyat `ember-cli-rails` gem has been updated as well in the second commit.
Ember, Ember data and the `ember-cli-rails-addon` have been updated during the ember-cli update.

vikhyat:
awesome

PR #372, opened 2015-01-15
Filthy feed posts

cybrox:
Allows users to mark feed posts as *nsfw*
Users with sfw-filter enabled won't see these posts.


cybrox:
Fatal git typo, awkward.

PR #373, opened 2015-01-15
Filthy feed posts [READY]

cybrox:
Allows users to mark feed posts as nsfw
Users with sfw-filter enabled won't see these posts.

Copy of #372

### DIFF
--- a/app/frontend/app/components/post-comment.js
+++ b/app/frontend/app/components/post-comment.js
@@ -1,23 +1,38 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-    classNames: ["status-update-panel"],
+  classNames: ["status-update-panel"],
+  formIsOpen: false,
 
   didInsertElement: function() {
     var self = this;
     this.$(".status-form").focus(function() {
-      self.$(".status-form").autosize({append: "\n"});
-      self.$(".panel-footer").slideDown(200);
+      self.set('formIsOpen', true);
+      self.handleFormState();
     });
     this.$(".status-form").blur(function() {
       if (self.$(".status-form").val().replace(/\s/g, '').length === 0) {
+        self.set('formIsOpen', false);
+        self.handleFormState();
+      }
+    });
+  },
+
+  handleFormState: function(){
+    var self = this;
+
+    setTimeout(function(){
+      if(self.get('formIsOpen')){
+        self.$(".status-form").autosize({append: "\n"});
+        self.$(".panel-footer").slideDown(200);
+      } else {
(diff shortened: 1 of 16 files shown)


app/frontend/app/templates/story/comment.hbs
```diff
@@ -52,6 +52,11 @@
 </div>
 
 <div class="story-actions">
+  {{#if adult}}
+    <span class="nswf-indicator" data-tooltip="This post has been marked as nsfw!">
```
NuckChorris: >nswf
>not safe work for

cybrox: Not safe wor fork!

vikhyat:
why is the build #failing?

cybrox:
I don't know, it just aborted the test without running into any errors.
I can't manually restart it though (afaik), you might want to do that.

app/frontend/app/components/post-comment.js
```diff
@@ -32,9 +50,18 @@ export default Ember.Component.extend({
...
   actions: {
     submitPost: function() {
-      var comment = this.get('newPost');
-      this.set('newPost', "");
-      this.sendAction('action', comment);
+      var comment = this.get('newPost'),
+          isAdult = this.get('newPostAdult');
+
+      this.setProperties({
+        'newPost': "",
```
vikhyat: Does the key need to be quoted here?

cybrox: No, it'll work either way though.
Should I change it?

vikhyat:
Looks good at first glance, will do a more thorough review tomorrow.

app/frontend/app/components/post-comment.js
```diff
@@ -1,23 +1,41 @@
...
+        self.$(".panel-footer").slideDown(200);
+      } else {
         self.$(".status-form").val('');
         self.$(".panel-footer").slideUp(200, function() {
           self.$(".status-form").trigger("autosize.destroy");
         });
       }
-    });
-  },
+    }, 300);
```
NuckChorris: What's up with the `setTimeout` here? Is it a workaround for a bug?  If so, you should probably add a comment explaining that

cybrox: It's a delay for the form blur so you can technically click the nsfw button before entering any text. I'll add a comment and squash.

NuckChorris:
Once that comment is shoved in, LGTM.  I don't see any potential blockers (though it might bring out some unnoticed bugs in the feed code)

PR #375, opened 2015-01-18
Migrate Waifus to be a character relation instead of a broken cache [WIP]

NuckChorris:
I wanted to get this out of my old heap of half-finished stuff and complete it, so I rebased my changes, fixed the N+1 with a more robust caching approach, and improved the typeahead

This solves a number of longstanding issues:
 * Waifu from Manga now properly link to the manga
 * Waifu will link to the earliest appearance of that character in any medium
 * It's impossible to get the `#` link waifu
 * The waifu is no longer a casting
 * The character ID is no longer stored in a zero-padded string which sometimes overflows to more characters than the 4 padded
 * It will display the source (first appearance's title) of a character in the autocomplete menu, since it's currently hell to find the right one if your waifu/husbando has a one-word name (like mine)

TODO:

 * [ ] Fix the typeahead and add the character's image
 * [x] Add a temporary redirect from `/character/:id` to their primary media, until we finish character pages, and remove the odd polymorphic `link-to`
 * [ ] Figure out polymorphic shit with Ember-Data (fix the build too)


app/serializers/user_serializer.rb
```diff
@@ -1,8 +1,10 @@
 class UserSerializer < ActiveModel::Serializer
   attributes :id, :cover_image_url, :avatar_template, :rating_type, :mini_bio,
     :about, :is_followed, :title_language_preference, :location, :website,
-    :waifu, :waifu_or_husbando, :waifu_slug, :waifu_char_id, :last_sign_in_at,
-    :current_sign_in_at, :is_admin, :following_count, :follower_count, :is_pro
+    :waifu_or_husbando, :last_sign_in_at, :current_sign_in_at, :is_admin,
+    :following_count, :follower_count, :is_pro
+
+  has_one :waifu, root: :characters
```
NuckChorris: This will cause an N+1 on pages where we load in multiple users, like the feeds.  Waifu should probably be moved to UserInfo.  Wasn't there something we were discussing about refactoring that? `User` and `FullUser` maybe?

PR #377, opened 2015-01-19
Remove margin-top on onboarding pages to reveal false header again

JoshFabian:


NuckChorris:
Are you sure you pushed the right changes?  Cause that commit removes the entire onboarding header, not the margin-top

NuckChorris:
In all fairness, you probably *did* fix the issue.

Just with a bit of collateral damage.

JoshFabian:
Yeah, I was uh, having issues with ember not displaying my updates so I tested just removing everything to see if it was caching or what. I forgot to revert before pushing. I'm dumb.

PR #381, opened 2015-01-20
Replace HasCurrentUser mixin with injection

NuckChorris:
We were basically mixing HasCurrentUser into everything anyways so I figured "let's just inject it into the controllers and get it over with"

This switches to an ObjectProxy-based service (soon to be an official role iirc) since ObjectController is deprecated and otherwise it caused recursion


vevix:
>we can probably reenable model injection in app.js now that the patch landed for that

As far as I'm aware, that was introduced in Ember 1.9+ and we are still on 1.8. I'm playing with updating Ember and its dependencies in a branch at the moment, so I can switch it there.

vikhyat:
`ObjectProxy` approach works well, I used that in a different Ember app a few months ago.

NuckChorris:
And finally, 3 months later, this is done and ready for review :dancer:

vevix:
I don't think you commited the services file

NuckChorris:
Balls. Brb lemme go do that

NuckChorris:
Fixed

vevix:
Looks good :+1:

Gonna need some updating with the Ember update

NuckChorris:
Yeah I figured it'd need some adjusting for that, but it should be mostly the same, besides having to rebase.  Probably a lot of merge conflicts.  I guess that image is quite fitting then, isn't it?

PR #386, opened 2015-01-21
fix library status reset bug

vevix:
ref: https://forums.hummingbird.me/t/changing-library-info-on-anime-pages-results-in-resets-and-errors/20471

Weird bug, where once we serialize the `LibraryEntry` with its `Anime` object from the server, the relationship between `FullAnime <-> LibraryEntry` gets set to null. ¯\\\_(ツ)_/¯

### DIFF
--- a/app/frontend/app/routes/anime/index.js
+++ b/app/frontend/app/routes/anime/index.js
@@ -25,7 +25,12 @@ export default Ember.Route.extend({
       return libraryEntry.save();
     }), {
       progressMessage: "Saving " + anime.get('canonicalTitle') + "...",
-      successMessage: "Saved " + anime.get('canonicalTitle') + "!"
+      successMessage: function() {
+        // update the 'full-anime' relationship, since it seems to
+        // disappear into the void
+        anime.set('libraryEntry', libraryEntry);
+        return "Saved " + anime.get('canonicalTitle') + "!";
+      }
     });
   },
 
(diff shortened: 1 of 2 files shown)


app/frontend/app/routes/anime/index.js
```diff
@@ -25,7 +25,12 @@ export default Ember.Route.extend({
       return libraryEntry.save();
     }), {
       progressMessage: "Saving " + anime.get('canonicalTitle') + "...",
-      successMessage: "Saved " + anime.get('canonicalTitle') + "!"
+      successMessage: function() {
```
NuckChorris: Does this problem also apply to Manga?

vevix: Most likely, I should add some Manga to my db.

vevix:
Applied to Manga as well :+1:

app/frontend/app/routes/anime/index.js
```diff
@@ -25,7 +25,12 @@ export default Ember.Route.extend({
...
+        // update the 'full-anime' relationship, since it seems to
+        // disappear into the void
+        anime.set('libraryEntry', libraryEntry);
+        return "Saved " + anime.get('canonicalTitle') + "!";
+      }
```
cybrox: Shouldn't these be `.property()`'s?

vevix: nope, they are functions invoked by actions.

cybrox: Oh, looked like a top level property in the diff snippet.
Just checked the complete file, you're right.
Looks good to me then.

PR #391, opened 2015-01-22
Default adult content to false

fractalemagic:
Tenpenchii doesn't provide this field so the value ends up being set as `null` which effectively marks it as adult content.


vikhyat:
LGTM

NuckChorris:
:+1:

PR #392, opened 2015-01-22
Favorite refactoring, final tweaks

cybrox:
Fixes all reported bugs, optimizes requests and adds a loading state

### DIFF
--- a/app/frontend/app/controllers/user/index.js
+++ b/app/frontend/app/controllers/user/index.js
@@ -37,6 +37,8 @@ export default Ember.ArrayController.extend(HasCurrentUser, {
   selectedWaifu: null,
   favoriteAnimePage: 1,
   favoriteMangaPage: 1,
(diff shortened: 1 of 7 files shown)


NuckChorris:
You should probably put more of this in the component than you are but for now it's a +1 from me

cybrox:
Well, I mainly focused the structures on just putting reusable things into components - or things we actually use elsewhere - so a lot of it is left in the user.index controller as before, yeah.
We might need to clarify, if we want to strictly follow the *component-ize everything* approcha in the future.

NuckChorris:
I think "componentize" should at least happen when you have nearly identical code between manga and anime :stuck_out_tongue:

cybrox:
Guess you're right on that one.
I'll just merge this to deploy the fixes for now and put the small *refactor thigns into the component* on the to-do list.

NuckChorris:
:+1:

PR #394, opened 2015-01-22
Worker fixes

NuckChorris:
Moved everything in the billing workers to be relative to the `last` and `current` parameters, which should fix all the fucking typecast errors, and make them more robust.

Also includes a hotfix for the MAL import worker, which just disables retries for now (they're not helping and are actually breaking things!)


vikhyat:
LGTM, can you merge this so I can go ahead and deploy?

PR #395, opened 2015-01-23
fix routing bug on favourites

vevix:
When navigating to an anime page of a users favourited anime, it results in the page not loading with an error of: `Cannot read property 'NaN' of undefined`.

This is because it is attempting to use data that exists on a 'full-anime' object and not the current 'anime' object.


cybrox:
Oh, I totally forgot about this bug.
Should definitely work, thanks for the fix :stuck_out_tongue:

PR #396, opened 2015-01-24
Collapse NSFW stories by default

fractalemagic:
It's not great, but it does the job for now.

Example

### DIFF
--- a/app/frontend/app/controllers/story.js
+++ b/app/frontend/app/controllers/story.js
@@ -16,6 +16,7 @@ export default Ember.ObjectController.extend(HasCurrentUser, {
   overflowing: false,
   showMoreText: 'Show More',
 
+  collapsed: Ember.computed.equal('model.adult', true),
   showAll: false,
   loadingAll: false,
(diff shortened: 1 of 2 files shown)


NuckChorris:
Few problems: first of all, this collapses all posts, doesn't it?

Second, we need to also collapse comment images.

We should probably use the same hooks we us for spoilers.

fractalemagic:
Oh right, I forgot about comments ¯\\_(ツ)_/¯ And no it won't collapse all posts, only adult ones.

`collapsed: Ember.computed.equal('model.adult', true)`

vevix:
Do you plan to update this to handle comments too? Otherwise, I'm a fan :+1:

/cc @Josh

JoshFabian:
This looks good to me! We can update it later to be a little sexier but I think this is a great start that gets the job done. Supporting comments would be ideal but something is better than nothing.

fractalemagic:
Comments are hidden :smiley: 

Hidden Comments

JoshFabian:
You rock

PR #397, opened 2015-01-24
Final patch for PRO Sidetiq

NuckChorris:
Despite the implication in the sparse Sidetiq documentation, they pass floats, not Time objects.  Casting fixes all issues.  Tested thoroughly.

Was gonna include a final patch for the MAL Import with this but I'd rather get this in sooner.


vevix:
:+1: :shipit:

PR #398, opened 2015-01-24
Not-really-minor-anymore fix for MAL import

NuckChorris:
DOES NOT fix the problem that was causing infinite retries; I have so far been unable to reproduce that issue locally.

But it *does* fix the constant empty-string imports.  Some (evidently quite common) combination of browser and operating system results in the `.xml.gz` file being uploaded as `.xml` — the best solution is to see if it's got the gzip magic number (`1F 8B 08`).  Another option is to use [Paperclip::ContentTypeDetector](http://www.rubydoc.info/github/thoughtbot/paperclip/Paperclip/ContentTypeDetector), which shells out to `file` and achieves the same thing.  It's a trade-off: we can either shell out or we can embed a magic number in our code.

It also switches to AJAX instead of the redirect-loop and provides a mechanism for getting errors out of the import system and back to the user.

Pivotal: https://www.pivotaltracker.com/story/show/91727174

### Todo:
 * [x] Update it in onboarding
 * [x] Toss a few files in the hole and test that it works right

### DIFF
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -34,9 +34,9 @@ class Settings::ImportController < ApplicationController
 
     mixpanel.track 'Imported from MyAnimeList', {email: current_user.email} if Rails.env.production?
   rescue Exception
-    error! 500, 'There was a problem importing your anime list.  Please send an
-                 email to josh@hummingbird.me with the file you are trying
-                 to import.'
+    error! 500, 'There was a problem importing your list.  Please send an email
+                 to josh@hummingbird.me with the file you are trying to import
+                 and we\'ll see what we can do.'
      raise
   end
 end
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -5,41 +5,35 @@ class Settings::ImportController < ApplicationController
   before_filter :authenticate_user!
 
   def myanimelist
-    hide_cover_image
+    params.require(:file)
 
-    if params[:animelist]
-      file = params[:animelist]
-      begin
-        # Check the magic number for gzip because some browsers are stupid
-        # Many users are uploading zipped-up lists with a text/xml MIME
-        if file.readpartial(3) == 0x1F_8B_08
-          gz = Zlib::GzipReader.new(file)
-          xml = gz.read
-          gz.close
-        elsif file.content_type.include?('xml')
-          xml = file.read
-        else
-          raise "Unknown format"
-        end
-
-        xml = XMLCleaner.clean(xml)
-
-        raise "Blank file" if xml.blank?
-
-        current_user.update_column :mal_import_in_progress, true
-        MyAnimeListImportApplyWorker.perform_async(current_user.id, xml)
-
-        mixpanel.track "Imported from MyAnimeList", {email: current_user.email} if Rails.env.production?
-      rescue Exception => e
-        flash[:error] = "There was a problem importing your anime list (#{e.message}).  Please send an email to <a href='mailto:vikhyat@hummingbird.me'>vikhyat@hummingbird.me</a> with the file you are trying to import.".html_safe
-        redirect_to "/users/edit"
-        return
-      end
+    # Prepare the file
+    # Check the magic number for gzip because some browsers are stupid
+    # Many users are uploading zipped-up lists with a text/xml MIME
+    if file.readpartial(3) == 0x1F_8B_08
+      gz = Zlib::GzipReader.new(file)
+      xml = gz.read
+      gz.close
+    elsif file.content_type.include?('xml')
+      xml = file.read
+    else
+      return error!(400, "Unknown format")
     end
+    xml = XMLCleaner.clean(xml)
 
-    unless current_user.mal_import_in_progress?
-      flash[:success] = "Import completed successfully."
-      redirect_to user_path(current_user)
-    end
+    return error!(422, "Blank file") if xml.blank?
+
+    # Queue the import
+    current_user.update_columns import_status: :queued,
+                                import_from: 'myanimelist'
+    MyAnimeListImportApplyWorker.perform_async(current_user.id, xml)
+
+    render json: current_user
+
+    mixpanel.track "Imported from MyAnimeList", {email: current_user.email} if Rails.env.production?
+  rescue Exception
+    error! 500, "There was a problem importing your anime list.  Please send an
+                 email to vikhyat@hummingbird.me with the file you are trying
+                 to import."
   end
 end
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -33,7 +33,7 @@ class Settings::ImportController < ApplicationController
     mixpanel.track "Imported from MyAnimeList", {email: current_user.email} if Rails.env.production?
   rescue Exception
     error! 500, "There was a problem importing your anime list.  Please send an
-                 email to vikhyat@hummingbird.me with the file you are trying
+                 email to josh@hummingbird.me with the file you are trying
                  to import."
   end
 end
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -8,23 +8,30 @@ class Settings::ImportController < ApplicationController
     hide_cover_image
 
     if params[:animelist]
+      file = params[:animelist]
       begin
-        if params[:animelist].content_type == "text/xml"
-          xml = params[:animelist].read
-        else
-          gz = Zlib::GzipReader.new(params[:animelist])
+        # Check the magic number for gzip because some browsers are stupid
+        # Many users are uploading zipped-up lists with a text/xml MIME
+        if file.readpartial(3) == 0x1F_8B_08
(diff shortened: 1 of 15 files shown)


vevix:
Is this WIP? So far the code is good :+1:

NuckChorris:
This can work on its own (and probably double the number of successful imports!), but there's a few other bugs that I'd *like* to address with it.

Namely that damn retry spam we were hitting.  I put a temp fix in place for that, but I'd really like something more permanent.

That being said, I really haven't been able to reproduce that locally :/ something about being unable to find

app/controllers/settings/import_controller.rb
```diff
@@ -8,23 +8,30 @@ def myanimelist
...
+        raise "Blank file" if xml.blank?
+
         current_user.update_column :mal_import_in_progress, true
         MyAnimeListImportApplyWorker.perform_async(current_user.id, xml)
 
         mixpanel.track "Imported from MyAnimeList", {email: current_user.email} if Rails.env.production?
-      rescue
-        flash[:error] = "There was a problem importing your anime list. Please send an email to <a href='mailto:vikhyat@hummingbird.me'>vikhyat@hummingbird.me</a> with the file you are trying to import.".html_safe
+      rescue Exception => e
+        flash[:error] = "There was a problem importing your anime list (#{e.message}).  Please send an email to <a href='mailto:vikhyat@hummingbird.me'>vikhyat@hummingbird.me</a> with the file you are trying to import.".html_safe
```
vikhyat: Revealing raw error messages to end users is not a good idea.

NuckChorris: I agree that it's not ideal to show them the raw error message, but we need to expose the most basic error info to the user somehow and this was just a quick hacked-together PR for avoiding (and helping debug) common failures.

That being said, it's also not as if a user will ever see this anyways: very few pages display the flash messages anymore.  I'm working on a better solution right now.

vikhyat: `Undefined method + for nil:NilClass` means nothing to our users, and doesn't mean anything to us either without a stacktrace.

NuckChorris: I'm aware of that but that's a rare case compared to the two errors I `raise`

app/workers/my_anime_list_import_apply_worker.rb
```diff
@@ -8,6 +8,9 @@ def perform(user_id, xml)
     malimport = MyAnimeListImport.new(user, xml)
     malimport.apply!
     user.recompute_life_spent_on_anime!
-    user.update_column :mal_import_in_progress, false
+    user.update_columns import_status: nil, import_from: nil
```
NuckChorris: This should probably also reset `import_error` actually

app/controllers/settings/import_controller.rb
```diff
@@ -5,34 +5,35 @@ class Settings::ImportController < ApplicationController
...
+    render json: current_user
 
-    unless current_user.mal_import_in_progress?
-      flash[:success] = "Import completed successfully."
-      redirect_to user_path(current_user)
-    end
+    mixpanel.track "Imported from MyAnimeList", {email: current_user.email} if Rails.env.production?
+  rescue Exception
+    error! 500, "There was a problem importing your anime list.  Please send an
+                 email to vikhyat@hummingbird.me with the file you are trying
```
NuckChorris: Should we have like... a mailing list address for stuff like this?  `ninetynine-problems@hummingbird.me` or smth?

JoshFabian: Nah, they should just email me.

NuckChorris:
I think it should be mostly functioning at this point.  I'm gonna throw some files in the hole tonight or tomorrow and see what happens.

DNA:
Seems good for me! :see_no_evil: :shipit:

NuckChorris:
Sadly it's very much not good.  Lots of mistakes throughout it, which result in it very very very not working

DNA:
At least it is a good improvement from before. Better a little improvement than no improvement.

And the main problem is in the importer that read that xml, isn't?

NuckChorris:
Actually it works worse than before, because it just _doesn't_ work lol

It's not done though, I'm gonna finish debugging tonight

NuckChorris:
We're all green for review on this; all the bugs seem to be squashed.

frontend/app/models/current-user.js
```diff
@@ -9,11 +10,50 @@ export default User.extend({
...
+    data.append('file', file);
+    return ajax('/settings/import/' + service, {
+      data: data,
+      cache: false,
+      contentType: false,
+      processData: false,
+      type: 'POST'
+    }).then(() => {
+      this.store.update('currentUser', {
+        id: this.get('id'),
```
vevix: Is this necessary?

frontend/app/models/current-user.js
```diff
@@ -9,11 +10,50 @@ export default User.extend({
...
+        importStatus: 'queued',
+        importingFrom: service
```
vevix: `importFrom`

frontend/app/templates/onboarding/library.hbs
```diff
@@ -12,10 +12,15 @@
...
   <div class="col-sm-12">
     <div class="import-panel">
-      {{#if importFromMal}}
-        <i class="fa fa-spinner"></i> Hang tight, we're importing your data.
+      {{#if currentUser.isImportErrored}}
+        <i class="fa fa-warning"></i> {{currentUser.importError}}
+      {{/if}}
+      {{#if currentUser.isImportOngoing}}
+        <i class="fa fa-spinner"></i>
+        We're importing your data.  Keep going, we'll tell you when it's done.
```
vevix: Double space :scissors:

frontend/app/models/current-user.js
NuckChorris: the `update` call is to update without dirtying the model, and I think it _requires_ the `id`

frontend/app/templates/onboarding/library.hbs
NuckChorris: You mean the actual spaces after the sentence right

frontend/app/models/current-user.js
NuckChorris: Naisu catchu!

frontend/app/templates/onboarding/library.hbs
vevix: Between `data.` and `Keep`

app/controllers/settings/import_controller.rb
```diff
@@ -5,34 +5,38 @@ class Settings::ImportController < ApplicationController
...
-    end
+    render json: current_user, serializer: CurrentUserSerializer
 
-    unless current_user.mal_import_in_progress?
-      flash[:success] = "Import completed successfully."
-      redirect_to user_path(current_user)
-    end
+    mixpanel.track 'Imported from MyAnimeList', {email: current_user.email} if Rails.env.production?
+  rescue Exception
+    error! 500, 'There was a problem importing your anime list.  Please send an
```
vevix: Don't we also import manga? Should probably drop it to "your list"

app/workers/my_anime_list_import_apply_worker.rb
```diff
@@ -8,6 +8,13 @@ def perform(user_id, xml)
     malimport = MyAnimeListImport.new(user, xml)
     malimport.apply!
     user.recompute_life_spent_on_anime!
-    user.update_column :mal_import_in_progress, false
+    user.update_columns import_status: nil, import_from: nil, import_error: nil
+  rescue Exception
+    status = User.import_statuses[:error]
+    user.update_columns import_status: status,
+      import_error: 'There was a problem importing your anime list.  Please
```
vevix: Same as above comment.

frontend/app/models/current-user.js
```diff
@@ -9,11 +10,50 @@ export default User.extend({
...
+      });
+    }, (err) => {
+      try {
+        this.store.update('currentUser', {
+          id: this.get('id'),
+          importStatus: 'error',
+          importError: err.jqXHR.responseJSON.error
+        });
+      } catch (e) {
+        alert("An unknown error occurred while attempting to import your list. Send an email to josh@hummingbird.me and we'll try to make it right");
```
vevix: Maybe this should match the error from the backend? Not sure.

vevix:
:+1: LGTM for when comments have been handled. :tada:

frontend/app/models/current-user.js
vevix: https://github.com/emberjs/data/blob/master/CHANGELOG.md#storeupdate-has-been-deprecated

frontend/app/models/current-user.js
```diff
@@ -9,11 +10,50 @@ export default User.extend({
...
+      type: 'POST'
+    }).then(() => {
+      this.store.update('currentUser', {
+        id: this.get('id'),
+        importStatus: 'queued',
+        importingFrom: service
+      });
+    }, (err) => {
+      try {
+        this.store.update('currentUser', {
```
vevix: Same as deprecation comment above.

https://github.com/hummingbird-me/hummingbird/pull/398/files#r34417670

NuckChorris:
All good?

PR #399, opened 2015-01-24
Improve the review summary partial.

vikhyat:
* x of y users found this review helpful.
* Avatar and username link to the user's page.


NuckChorris:
:+1: on the code but how's it look?  I worry that it would be rather cluttered to have the votes in the bubble?

PR #401, opened 2015-01-28
Make QU background transition smoother

NuckChorris:
Uses a gradient position change instead of an opacity change, so that the background appears to fade in along with the button and title

I just kind of got frustrated with how the background suddenly turns black but the other things fade in and whipped this out real fast in dev tools :D 

Obligatory gif:


JoshFabian:
LGTM

PR #402, opened 2015-01-29
Publish manga favs and nsfw posts

cybrox:
- Publish manga favorites and nsfw posts for all users
- Hide Manga box for other users if it's empty (prevent ugly double error message)

### DIFF
--- a/app/frontend/app/components/favorite-container.js
+++ b/app/frontend/app/components/favorite-container.js
@@ -12,6 +12,7 @@ export default Ember.Component.extend({
   isEditing: false,
   isLoading: true,
   showPages: 1,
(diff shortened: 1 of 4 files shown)


NuckChorris:
Outside of the showSelf, this is a +1 from me, let's get that shit to the masses!

cybrox:
I guess using `showContainer` would be a solution?

NuckChorris:
Would `showMangaFavorites` be better?  It's kind of redundant due to context but it certainly gets the point across

cybrox:
Yeah, I guess that'd make sense.

NuckChorris:
Once the build light stops eyeing us judgementally, lgtm

PR #403, opened 2015-01-29
Fix the whitespace above the Community box

NuckChorris:
Margins collapsing caused it to go boom in some browsers.

Floating left seems to fix this.  Might break other things.

#nuckfixestinynuisanceswhiletired


vikhyat:
yay!

PR #407, opened 2015-02-03
Fix avatar shrinking before break into mobile layout

Wopian:
Just getting rid of remnants from the old user page design.

Before:

After:


JoshFabian:
I think some shrinkage is still needed so that the avatar isn't so tight up against the sub-nav.

Wopian:
:+1: reverted to 109px earlier, which makes the avatar in-line with the bottom of the bio text.

Wopian:
Updated `After` image to reflect changes @JoshFabian

JoshFabian:
LGTM! +1

PR #408, opened 2015-02-04
Fix quote submit routine for anime

cybrox:
This will fix submitting quotes for animes on the `/anime/<id>/quotes` page.
The whole form might need a spam protection for the future as well or a moderative staging layer.

### DIFF
--- a/app/controllers/quotes_controller.rb
+++ b/app/controllers/quotes_controller.rb
@@ -16,15 +16,18 @@ class QuotesController < ApplicationController
 
   def create
     authenticate_user!
-    @anime = Anime.find(params["quote"]["anime_id"])
-    @quote = Quote.new
-    @quote.content = params["quote"]["content"]
-    @quote.character_name = params["quote"]["character_name"]
-    @quote.anime = @anime
-    @quote.user = current_user
-    @quote.save
 
-    redirect_to anime_quotes_path(@anime)
+    quote_params = params.require(:quote).permit(:anime_id, :character_name, :content)
+
+    anime = Anime.find(quote_params[:anime_id])
+    quote = Quote.create(
+      character_name: quote_params[:character_name],
+      content: quote_params[:content],
+      anime: anime,
+      user: current_user
+    )
+
+    render json: quote
   end
 
   def update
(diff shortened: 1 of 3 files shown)


app/controllers/quotes_controller.rb
```diff
@@ -16,15 +16,20 @@ def index
...
-    @quote.content = params["quote"]["content"]
-    @quote.character_name = params["quote"]["character_name"]
-    @quote.anime = @anime
-    @quote.user = current_user
-    @quote.save
-
-    redirect_to anime_quotes_path(@anime)
+
+    quote_params = params.require(:quote).permit(
+      :anime_id, :character_name, :content, :favorite_count, :username
```
NuckChorris: Uh.  Should the user really be able to set the `favorite_count`? And should `username` be in here too?  I think you might be abusing `permit`.

cybrox: Well, it wouldn't make a difference since the data is not written to the model but I should probably remove that from there after copying the params from request :frowning: 
Will fix and squas.

app/frontend/app/controllers/anime/quotes.js
```diff
@@ -37,6 +30,7 @@ export default Ember.ArrayController.extend(HasCurrentUser, {
           });
 
       quote.save();
+      this.get('content').addObject(quote);
```
NuckChorris: Shouldn't ED handle this automatically via data binding to the model array?

cybrox: No, creating and saving the record will push it to the store. While this might update every property dependent on that via `model.@each`, it does not update the route model unless you actively reload it. - As far as I'm aware of.

app/controllers/quotes_controller.rb
NuckChorris: If you want a generated hash out of it you can always `params.require(:quote).permit(:anime_id).to_h` though the keys will be stringified (for some reason that I still don't understand)

app/frontend/app/controllers/anime/quotes.js
NuckChorris: hmm.  I don't know enough to say one way or the other, so I'll just assume you're right

cybrox: I just tested it without that line, it doesn't get automatically pushed to the model. I guess an alternative way would be using `model.refresh();` but that's triggering all the model hooks again so I think this is the best solution for handling a simple object-push.

NuckChorris:
:+1: LGTM

PR #409, opened 2015-02-06
Groups

NuckChorris:
Since the checklist seems to be done, let's get reviewing this monster.

In hindsight, it was probably a bad idea to do this in one big feature branch.


app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    if membership.rank_changed? && membership.rank_was == 'admin' && !group.can_admin_resign?
+      return error! "Last admin of the group cannot resign", 400
+    else
+      membership.save!
+      render json: membership
+    end
+  end
+
+  def destroy
+    return error! "Mods can only remove regular users from the group", 403 unless current_user.admin? || current_member.admin? || (current_member.mod? && membership.pleb?) || current_user.id == membership.user_id
```
NuckChorris: If somebody wants to make this conditional less shit it'd be welcome

app/frontend/app/controllers/story.js
```diff
@@ -21,6 +21,14 @@ export default Ember.ObjectController.extend(HasCurrentUser, {
   loadingAll: false,
   loadedAll: propertyEqual('substories.length', 'model.substoryCount'),
 
+  // Determine if we are on the stories group page
+  // this is really hacky
+  isOnGroupPage: function() {
+    // 8 == /groups/
+    return this.get('model.group') &&
+      (this.get('target.target.url').substr(8) === this.get('model.group.id'));
```
NuckChorris: Do we have to do such a complex thing, can't we just check if the url just has "groups" in it?  I mean, it's not like posts to "Hello Kitty Fanclub" will be on pages for "Nuck's Yaoi Harem", right?

app/frontend/app/models/group-member.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default DS.Model.extend({
+  user: DS.belongsTo('user'),
+  rank: DS.attr('string'),
+  pending: DS.attr('boolean'),
+  groupId: DS.attr('string'),
+
+  // This suffers the same problem as `substories`, where the
+  // relationship only serializes a subset of `group-members`,
+  // rather than all and ember will destroy the relationship.
```
NuckChorris: Is this a bug, should we report it, or is it already reported?  If it's already reported, we should put a link to the issue so down the road we can check the status and potentially remove the hack

app/frontend/app/controllers/story.js
cybrox: The ApplicationController holds `currentPaht` and `currentRouteName`, that should get this done a bit less hackier.

app/frontend/app/models/story.js
```diff
@@ -30,6 +31,8 @@ export default Model.extend(ModelCurrentUser, {
   }.property('poster.id', 'user.id'),
 
   canDeleteStory: function() {
-    return (!this.get('isNew')) && (this.get('belongsToUser') || this.get('currentUser.isAdmin'));
-  }.property('isNew', 'belongsToUser', 'currentUser.isAdmin'),
+    var groupMember = this.get('group') && this.get('group').isMember(this.get('currentUser'));
+    return (!this.get('isNew')) && (this.get('belongsToUser') || this.get('currentUser.isAdmin') ||
+      (groupMember && (groupMember.get('isAdmin') || groupMember.get('isMod'))));
+  }.property('isNew', 'belongsToUser', 'currentUser.isAdmin', 'group'),
```
NuckChorris: My soul aches from this conditional.  Can we split it onto multiple lines and also like `if this.get('currentUser.isAdmin') return true;` to emphasize the fact that admin is short-circuited true?

app/frontend/app/routes/group/manage.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+
+  setupController: function(controller, model) {
+    // redirect users to group page if they aren't admins
+    if (!controller.get('currentMember.isAdmin')) {
+      this.transitionTo('group', model.get('id'));
+    }
+    controller.set('model', model);
+  }
+});
```
NuckChorris: I think we nuked the group management page so this (and the other manage stuff) should probably get pulled

app/frontend/app/templates/group.hbs
```diff
@@ -0,0 +1,146 @@
...
+{{! Ad Banner }}
+<div class="container">
+  <div class="feed-head-advert">
+    {{ad-unit adId="1293413" adClass="257f81e798bd68dd81e60f42838f361f"}}
+  </div>
+</div>
+
+{{! Content }}
+<div class="min-height-wrapper">
+  {{#if model.closed}}
```
NuckChorris: Needs removed, since group closure is no longer a thing

app/frontend/app/routes/group/index.js
```diff
@@ -0,0 +1,51 @@
...
+  setupController: function(controller, model) {
+    setTitle(controller.get('group.name'));
+
+    var groups = this.store.find('group', {
+      limit: 3
+    });
+    controller.set('suggestedGroups', groups);
+
+    this.setCanLoadMore(true);
+    controller.set('canLoadMore', true);
```
cybrox: Could be wrapped into one `setProperties` but ¯\_(ツ)_/¯

app/frontend/app/models/group-member.js
vevix: It's not really a bug with Ember, just the way we have our API setup.

Because we only serialize a small amount of the members on the `GroupSerializer`, when we add more to that relation via visiting the `/members` page, Ember destroys all of those new relations when it gets that same `Group` data from another JSON payload.

There is probably a better way to handle it but this is the same hack I applied to substories (to fix the deletion bug).

app/frontend/app/routes/group/manage.js
vevix: I think the manage page is still relevant for an admin closing their group, unless we want to move that functionality elsewhere?

app/frontend/app/templates/group/members.hbs
```diff
@@ -0,0 +1,142 @@
+<div class="container">
+
+  <!-- Only show pending section if there are pending members -->
+  {{#if hasPendings}}
```
NuckChorris: We should probably only show pending users to mods and admins

vevix: That is how it is AFAIK

app/frontend/app/routes/group/manage.js
NuckChorris: Ah, so we should only expose it to `{{#if currentUser.isAdmin}}` then?  That works

vevix: AFAIK the Manage page only shows up to group admins, and has redirection in the route for anyone trying to bypass.

app/frontend/app/templates/group/manage.hbs
```diff
@@ -0,0 +1,20 @@
...
+          <div class="panel-heading">
+            Group Settings
+          </div>
+          <div class="panel-body">
+            {{#if model.closed}}
+              Your group is closed! You'll need to contact a staff member to re-open it.
+            {{else}}
+              <button class="btn btn-danger" {{action "closeGroup"}}>
+                Close Group
+              </button>
```
NuckChorris: If we're gonna maintain this page for site administrators then we should rephrase it to "Delete Group and all posts seriously don't fucking do this unless the group is evil"

app/frontend/app/templates/group.hbs
vevix: If group closure has been removed then there is a ton of `{{group.closed}}` conditionals around in both the Handlebar templates, and Javascript that will need to be removed.

app/frontend/app/templates/group/manage.hbs
cybrox: Shouldn't this be removed anyways? I though this action would be restricted to `currentUser.isAdmin` instead of `currentMember.isAdmin` now.

app/frontend/app/routes/groups.js
```diff
@@ -0,0 +1,26 @@
...
+    return this.store.find('group', {
+      trending: true,
+      page: page
+    });
+  },
+
+  setupController: function(controller, model) {
+    this.setCanLoadMore(true);
+    controller.set('model', model);
+    controller.set('recentGroups', this.store.find('group'));
```
cybrox: `setProperties`.

app/frontend/app/templates/group.hbs
NuckChorris: I removed group closure in efb6bef as josh and I had discussed in chat previously.  I think the general sentiment is that putting group success at the whim of the group admins is risky.  If a user wants to close it, they can kick all the users out and ask us to delete it for them.

Original plan was to give 30 days of grace where the user could un-close it, but once it's out of the hands of users I just opted for `.delay.destroy`

vevix:
"View Group" button state needs to be implemented on the GroupBox component IIRC.

cybrox:
Just wanted to ask about that, I'll quickly add that then.

app/models/story.rb
```diff
@@ -30,6 +32,7 @@ class Story < ActiveRecord::Base
   has_many :notifications, as: :source, dependent: :destroy
 
   validates :user, :story_type, presence: true
+  validates :group, presence: true, if: Proc.new {|s| s.group_id.present? }
```
NuckChorris: I am so confused, doesn't this literally mean "if group is present, check that group is present"?

vevix: If the ID is present then it validates the relationship, as in it's an actual valid object.

app/frontend/app/templates/group/manage.hbs
NuckChorris: It will be restricted, but the question is whether we ought to maintain this page.  I'm thinking we should just swap it for a "delete" button on the group page and then `if (prompt('Type the group slug to confirm') === this.get('model.id')) this.get('model').destroy();` or something — we can totally get away with using `prompt` since it's administrative stuff.

app/assets/stylesheets/components/community-widget.css.scss
```diff
@@ -178,3 +178,17 @@
   0% { background-position: -468px 0; }
   100% { background-position: 468px 0; }
 }
+
+.trending-group-panel {
```
vikhyat: Does this really belong in the community widget CSS file?

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,72 @@
+class GroupMembersController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    group = Group.find(params[:group_id])
+
+    res = current_user &&
```
vikhyat: Variable name needs to be more descriptive.

test/fixtures/groups.yml
```diff
@@ -0,0 +1,33 @@
...
+#  cover_image_file_size    :integer
+#  cover_image_updated_at   :datetime
+#  confirmed_members_count  :integer          default(0)
+#  created_at               :datetime
+#  updated_at               :datetime
+#  avatar_processing        :boolean
+#
+
+gumi:
+  name: Gumi Appreciation Group
```
vikhyat: :+1:

app/frontend/app/templates/group/members.hbs
vikhyat: `hasPendingMembers` would be a better name.

app/models/group.rb
```diff
@@ -0,0 +1,96 @@
...
+  extend FriendlyId
+  friendly_id :name, :use => [:slugged, :history]
+
+  # We don't need to callback if we're killing the group off, just delete all
+  # memberships.
+  has_many :members, foreign_key: :group_id,
+    class_name: 'GroupMember', dependent: :delete_all
+  has_many :stories
+
+  has_attached_file :avatar,
```
vikhyat: Let's use a better URL scheme for our newer attachments, something like `/group_avatar/:group_id.:ext` would be a lot better than the paperclip default with ID partitioning and original filenames being preserved.

app/models/group_member.rb
```diff
@@ -0,0 +1,44 @@
...
+    update_group_scores!
+    update_counter_cache!
+  end
+
+  after_destroy do
+    update_counter_cache!
+  end
+
+  def update_counter_cache!
+    self.group.update_attribute :confirmed_members_count,
```
vikhyat: Do we need to do a full count here? Can't we do atomic increment/decrement?

app/services/trending_groups.rb
```diff
@@ -0,0 +1,26 @@
+class TrendingGroups
```
vikhyat: We shouldn't be duplicating code between TrendingAnime and TrendingGroups like this. Extract the common functionality into a common parent class or something.

db/migrate/20141230222400_create_groups.rb
```diff
@@ -0,0 +1,17 @@
+class CreateGroups < ActiveRecord::Migration
+  def change
+    create_table :groups do |t|
+      t.string :name, unique: true, null: false
```
vikhyat: Do we need the ability to search by name?

db/migrate/20141230222755_create_group_members.rb
```diff
@@ -0,0 +1,13 @@
+class CreateGroupMembers < ActiveRecord::Migration
+  def change
+    create_table :group_members do |t|
+      t.references :user, index: true, null: false
+      t.references :group, index: true, null: false
+      t.boolean :admin, null: false, default: false
+      t.boolean :pending, null: false, default: true
+
+      t.index [:user_id, :group_id], unique: true
```
vikhyat: We have excessive index coverage here, having indexes on `user`, `group` and `user, group` is redundant. We need the combined index to enforce uniqueness so probably get rid of the index on user or group.

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    members = res ? group.members : group.members.accepted
+    members = members.page(params[:page]).per(20)
+    render json: members, meta: {cursor: 1 + (params[:page] || 1).to_i}
+  end
+
+  def create
+    membership_hash = params.require(:group_member).permit(:group_id, :user_id).to_h
+    user = User.find(membership_hash['user_id'])
+
+    return error! "Wrong user", 403 if user.nil? || current_user != user
```
vikhyat: Is it even possible for the user to be nil? I thought `#find` raises an error if the record is not found. And if the user must be the current user why even get the user ID from the client?

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,72 @@
+class GroupMembersController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    group = Group.find(params[:group_id])
+
+    res = current_user &&
+      (group.is_admin?(current_user) || group.is_mod?(current_user))
```
vikhyat: We're using `current_member.admin?` in some places and `group.is_admin?(current_user)` in others. Should make this consistent.

app/models/group.rb
```diff
@@ -0,0 +1,96 @@
...
+    presence: true,
+    uniqueness: {case_sensitive: false},
+    length: {minimum: 3, maximum: 30}
+
+  def can_admin_resign?
+    members.admin.count > 1
+  end
+
+  def member(user)
+    members.where(user: user).first
```
vikhyat: `members.find_by(user_id: user.id)`

app/models/group.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def is_admin?(user)
```
vikhyat: `is_predicate?` is not idiomatic Ruby, change these to `predicate?`.

app/models/group.rb
```diff
@@ -0,0 +1,96 @@
...
+    group = Group.new(params)
+    group.members.build(
+      user: admin,
+      rank: :admin,
+      pending: false
+    )
+    group
+  end
+
+  def self.trending(count = 10)
```
vikhyat: Do we really need this method?

app/services/news_feed.rb
```diff
@@ -65,8 +66,13 @@ def fetch(page=nil)
   # Regenerate the user's feed from scratch.
   def regenerate_feed!
     user_set = active_followed_users + [@user.id]
+    group_set = active_joined_groups.to_a
```
vikhyat: Is the `to_a` needed?

app/services/news_feed.rb
```diff
@@ -34,6 +34,7 @@ class NewsFeed
   USER_FOLLOWERS_PREFIX         = "user_followers:"
   USER_FOLLOWING_PREFIX         = "user_following:"
   ACTIVE_FOLLOWED_USERS_PREFIX  = "active_followed_users:"
+  ACTIVE_JOINED_GROUPS_PREFIX   = "active_joined_groups:"
```
vikhyat: We don't need these keys anymore since we're not using Redis as a cache for this information now.

app/assets/stylesheets/partials/groups.css.scss
```diff
@@ -0,0 +1,139 @@
...
+
+.group-kick, .group-reject {
+  padding: 5px 15px;
+  font-size: 12px;
+  background: #c0392b;
+  border: none;
+}
+
+// groups page
+
```
vikhyat: :scissors:

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    members = res ? group.members : group.members.accepted
+    members = members.page(params[:page]).per(20)
+    render json: members, meta: {cursor: 1 + (params[:page] || 1).to_i}
+  end
+
+  def create
+    membership_hash = params.require(:group_member).permit(:group_id, :user_id).to_h
```
vikhyat: Do we need the `to_h` here? I'm guessing `user_id` on the next line is a string instead of a symbol because of this?

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,85 @@
+class GroupsController < ApplicationController
+  def index
+    # currently also using this endpoint to pull down suggested groups
+    groups = if params[:limit].present? # Get a specific amount of trending
+      Group.trending(params[:limit].to_i)
+    elsif params[:trending].present? # Get all trending groups, paginated
+      Group.trending(0).page(params[:page]).per(20)
```
vikhyat: Instead of passing 0 to `#trending` we should support calling it without arguments.

vikhyat: Also I don't believe this will actually work.

`Group.trending(0)` will do a Redis lookup that returns the IDs of _all_ the trending groups. Then we do a `Group.where(id: <ids>)` which can return results in any order.

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,85 @@
+class GroupsController < ApplicationController
+  def index
+    # currently also using this endpoint to pull down suggested groups
+    groups = if params[:limit].present? # Get a specific amount of trending
+      Group.trending(params[:limit].to_i)
```
vikhyat: We should have an upper bound on this, don't want to allow people to cause us to serialize 1000 groups.

app/frontend/app/components/group-box.js
```diff
@@ -0,0 +1,51 @@
+import Ember from 'ember';
+/* global Messenger */
+
+export default Ember.Component.extend({
+
+  userIsMemberOfGroup: function(){
```
vikhyat: This is something that we should be serializiing from the server, not computing client side. See how we handle returning whether the current user is following someone for an example.

app/frontend/app/controllers/group.js
```diff
@@ -0,0 +1,99 @@
...
+  // to "Join Group" when waiting for the server to respond.
+  loading: false,
+
+  coverUpload: Ember.Object.create(),
+  coverUrl: Ember.computed.any('coverUpload.croppedImage', 'model.coverImage'),
+  coverImageStyle: function() {
+    return "background-image: url(" + this.get('coverUrl') + ")";
+  }.property('coverUrl'),
+
+  currentMember: function() {
```
vikhyat: This shouldn't depend on us having a complete list of all members of a group on the client. There can be groups with a very large number of members.

app/frontend/app/controllers/group.js
```diff
@@ -0,0 +1,99 @@
+import Ember from 'ember';
+import HasCurrentUser from '../mixins/has-current-user';
+/* global Messenger */
+
+export default Ember.Controller.extend(HasCurrentUser, {
+  // this flag is used so that the "Leave Group" button doesn't switch
+  // to "Join Group" when waiting for the server to respond.
+  loading: false,
+
+  coverUpload: Ember.Object.create(),
```
vikhyat: Let's not duplicate the cover upload code, extract it from here and users into a mixin.

app/frontend/app/controllers/group.js
```diff
@@ -0,0 +1,99 @@
+import Ember from 'ember';
+import HasCurrentUser from '../mixins/has-current-user';
+/* global Messenger */
+
+export default Ember.Controller.extend(HasCurrentUser, {
+  // this flag is used so that the "Leave Group" button doesn't switch
+  // to "Join Group" when waiting for the server to respond.
+  loading: false,
```
vikhyat: Why shouldn't the button switch while waiting for a server response?

app/frontend/app/controllers/group.js
```diff
@@ -0,0 +1,99 @@
...
+  showEditMenu: false,
+
+  actions: {
+    joinGroup: function(group) {
+      var member = this.store.createRecord('group-member', {
+        groupId: group.get('id'),
+        user: this.get('currentUser.model.content'),
+        pending: true
+      });
+      Messenger().expectPromise(function() {
```
vikhyat: Can't we do `member.save` instead of `function() { return member.save(); }`?

app/services/news_feed.rb
vevix: It was throwing a PG error without the conversion, Can't recall the error exactly.

app/frontend/app/controllers/modals/create-group.js
```diff
@@ -0,0 +1,43 @@
...
+import ModalMixin from '../../mixins/modals/controller';
+/* global Messenger */
+
+export default Ember.Controller.extend(ModalMixin, {
+  name: null,
+  bio: null,
+  about: null,
+
+  isDisabled: function() {
+    return (this.get('name') === null || this.get('name').length === 0) ||
```
vikhyat: `Ember.isBlank`

app/frontend/app/controllers/story.js
```diff
@@ -51,6 +59,14 @@ export default Ember.ObjectController.extend(HasCurrentUser, {
 
       var self = this;
       this.store.find('user', this.get('currentUser.id')).then(function(user) {
+        // group permission check
```
vikhyat: Why even allow this event to be triggered? If I'm not a member I should  not be shown the form.

app/controllers/groups_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    authenticate_user!
+
+    # Only site admins should be able to delete
+    # Sorry users, once you make a group it's part of the commons.
+    if current_user.admin?
+      group = Group.find(params[:id])
+      group.delay.destroy
+      render json: {}
+    else
+      return error! "Only hummingbird admininistrators can delete groups", 403
```
vikhyat: Hummingbird as a proper noun pls.

app/frontend/app/components/group-box.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    if(user.length !== 1) { return false; }
+    return user[0].get('pending');
+  }.property('userIsMemberOfGroup'),
+
+
+  actions: {
+    joinGroup: function(){
+      var member = this.get('targetObject.store').createRecord('group-member', {
```
vikhyat: Components should not access the store like this. Bubble an event upstream.

app/frontend/app/controllers/story.js
vikhyat: We need to measure the performance impact before adding `needs: ["application"]` here. This controller is super important for our story render times.

app/frontend/app/components/group-box.js
cybrox: To be honest, I don't really see why that would make sense for completely independent components. As far as I know, Ember wants to go the *componentize everything* way in the future, which makes perfect sense but assuming we have a component like this, that is used on a lot of different pages, it doesn't really make sense to put that action in every controller and just putting it up in the application controller is messy as well.
So the best solution would be to require the respective controller from every other controller with said action then and bubble it to that? - Kinda laborious.

app/frontend/app/models/group.js
```diff
@@ -0,0 +1,47 @@
...
+    var members = this.get('members');
+    return members.filterBy('pending', false).slice(0, 14);
+  }.property('members.@each'),
+
+  isMember: function(user) {
+    return this.get('members').filterBy('pending', false)
+      .findBy('user.id', user.get('id'));
+  },
+
+  truncatedBio: function(key, value) {
```
vikhyat: Looks like this is also duplicated in the user model and here.

app/frontend/app/routes/group.js
```diff
@@ -0,0 +1,25 @@
...
+
+  actions: {
+    closeModal: function() {
+      this.controller.set('coverUpload', Ember.Object.create());
+      return true;
+    },
+
+    uploadCover: function(image) {
+      this.currentModel.set('coverImage', image);
+      return Messenger().expectPromise(() => {
```
vikhyat: Again this could've just been `this.currentModel.save`?

app/controllers/group_members_controller.rb
vevix: `current_user` is used here as the `current_member` method is based on a GroupMember ID being passed as a param.

app/controllers/group_members_controller.rb
NuckChorris: Ember Data does it so I do it.

app/controllers/group_members_controller.rb
NuckChorris: The reason is (iirc, it's been a few weeks) because of cases where the params are used in unsafe ways — to_h clears out all keys besides those allowed. I'll check this more in morning.

app/frontend/app/controllers/group.js
NuckChorris: I do t think it does. Thomas made it serialize 14 + self, which means the self will always be present if the group is present

app/models/group.rb
NuckChorris: The partitioning I believe is to avoid hundreds of thousands of files per directory (some filesystems choke on that) by limiting it to 999 per. Is this an issue on S3? Agree that that's a better scheme to remove original name though

db/migrate/20141230222755_create_group_members.rb
NuckChorris: But we do lookups individually on both user and group id quite often, why wouldn't we have indexes on the two main lookup keys?

vikhyat: http://www.monkeyatlarge.com/archives/2011/02/08/redundant-indexing-in-postgresql/

app/models/group.rb
vikhyat: I am aware of why it does the ID partitioning, thank you very much. S3 doesn't have a hierarchical namespace, even if the key contains "/" it is just treated like a regular string.

app/frontend/app/controllers/group.js
vevix: Just thought it gave a better experience. 

When the server returned an error, I thought the switching of "Leave Group" -> "Join Group" -> "Leave Group" was bad.

app/frontend/app/controllers/group.js
vevix: Unfortunately not, Messenger calls `.apply` on the promise passed, which overwrites the `this` pointer and the `save()` call fails.

app/services/trending_groups.rb
NuckChorris: I was gonna do that but there was something that got in my way.  I'll look into it again.

db/migrate/20141230222400_create_groups.rb
NuckChorris: Ah, yes we should probably add the pg_search stuff

app/controllers/group_members_controller.rb
NuckChorris: We could also express it as `group.member(current_user).admin?` but it'd probably be slower since AR would have to deserialize and marshal a whole Member object while (iirc) `is_admin?(current_user)` uses an `exists?` call

app/services/news_feed.rb
NuckChorris: We aren't?  I didn't even notice.  I'll fix this.

app/frontend/app/controllers/group.js
NuckChorris: Could we abuse fat arrows to shorten this to `() => member.save()`?  I don't even think the curlies (or return) are needed for fat arrows with single statements, I think they behave like Ruby's `{}` blocks

app/models/group.rb
NuckChorris: It's not necessary but I feel it's a good thing and belongs here because it provides a matching interface to similar things like scopes and hides the TrendingGroups class from the controllers layer (reducing surface area despite additional volume), two positive attributes weighed against three additional lines.

app/models/group_member.rb
NuckChorris: Atomic increment and decrement would probably be good, yeah.  I'll look into it.

cybrox:
@vevix As far as I can see, this could be solved without the _force a parameter in just to make it query_ by using [store.fetch](http://emberjs.com/api/data/classes/DS.Store.html#method_fetch)

vevix:
Isn't that single record only?

cybrox:
It is, I just noticed. I was so sure there was a bulk-force-reload method once.
Nvm then, will add the **{}** back.

>  maybe a better method is to pass a 'recent' param to the endpoint?
Wouldn't that change the output to trending groups instead of recent groups?

NuckChorris:
Bringing the todo list here:

* [ ] Ember: See if we can simplify `StoryController#isOnGroupPage`
* [ ] Ember: Clarify `Story#canDeleteStory`'s conditional
* [x] Ember: Nuke group management page and replace with button to delete group (for site admins)
* [x] Ember: Remove all remaining references to `group.closed`
* [x] CSS: Rename `community-widget.css.scss` to `dashboard-widgets.css.scss`
* [ ] Rails: Change Paperclip URL scheme to `/group_avatar/:group_id.:ext`
* [ ] Rails: Change `GroupMember#update_counter_cache!` to use atomic incr/decr
* [ ] Rails: Deduplicate Trending systems
* [ ] Rails: Add pg_search to groups
* [ ] Rails: Remove `group_members` index on `group_id`
* [x] Rails: `Group#is_mod?` to `Group#has_mod?`
* [ ] Rails: Remove the unused keys from `NewsFeed`
* [ ] IDK: Fix the trending thing https://github.com/hummingbird-me/hummingbird/pull/409#discussion_r24233523
* [ ] Rails: Upper bound on `GroupsController` pagination
* [x] Rails+Ember: Move `userIsMemberOfGroup` to serialize on the serverside
* [x] Ember: Extract cover image upload, truncatedBio into a mixin
* [x] Rails: `GroupsController#destroy` — hummingbird is a proper noun

app/controllers/groups_controller.rb
vevix: Isn't is always going to be in the order of the array of ids passed?

vikhyat: Nope.

NuckChorris: If that's not a valid assumption then `TrendingAnime.list` is broken and so is `TrendingGroups.list` by proxy.  How can we fulfill that assumption if it's false?

app/assets/stylesheets/partials/dashboard-widgets.scss
```diff
@@ -1,4 +1,4 @@
-.community-widget-panel {
+.community-widget-panel, .trending-group-panel {
```
NuckChorris: We should probably use a `.dashboard-widget-panel` class instead?

app/frontend/app/components/group-box.js
vikhyat: Okay, makes sense. I wonder if it would be a good idea to inject the store into components though, would make this a lot cleaner.

app/controllers/group_members_controller.rb
vikhyat: Why not just have `user = current_user` instead of lines 17 to 19?

NuckChorris: Idk, it just felt like an odd behavior to silently convert any random user ID to the current user, instead of throwing an error

app/controllers/group_members_controller.rb
```diff
@@ -0,0 +1,73 @@
+class GroupMembersController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    group = Group.find(params[:group_id])
+
+    userHasModPermissions = current_user &&
```
vikhyat: No camelCase in Ruby!

PR #410, opened 2015-02-08
Fix apps header

KustoSan:
Fixing apps header at width <=624

From: http://puu.sh/fDMUV/3000101f8b.jpg
To: http://puu.sh/fDMPl/879d68ae15.jpg

### DIFF
--- a/app/assets/stylesheets/partials/apps-page.css.scss
+++ b/app/assets/stylesheets/partials/apps-page.css.scss
@@ -4,6 +4,12 @@
   background-size: cover;
   color: #fff;
   padding-top: 50px;
+  .container {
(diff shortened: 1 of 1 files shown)


Wopian:
Looks like it doesn't fix layout on widths under `423px`. Also produces an undesired position between `537px` and `624px`

NuckChorris:
The problem is that the button is just flow'd into place.  It should probably be absolutely positioned in the header box.

Trying to adjust it at breakpoints is opening a nasty can of worms since text renders differently on different platforms.  Wrapping will be different everywhere.

vikhyat:
Thanks!

PR #413, opened 2015-02-16
Show redeemed codes when there aren't any codes left.

vevix:
Found a bug where a users redeemed code wasn't showing when there weren't any pro codes left. :scream:


NuckChorris:
+1 jolly good lgtm

PR #415, opened 2015-02-17
Fix QU gradient for Firefox (browser-side malformed CSS)

Wopian:
https://forums.hummingbird.me/t/dark-overlay-on-quick-update/21893

Firefox seems to change the property to `background-size: 200% auto` which is 100% what we don't want. Applying `background-size: auto 200%` prevents this behaviour and works on all browsers.

FireFox 35:

Chrome 42:


NuckChorris:
Firefox interprets the single-value version as just width, while I think Webkit interprets it as both height and width. This should fix it :+1:

PR #418, opened 2015-02-20
Group search & refactoring

NuckChorris:
Serverside changes:

 * Extract the presenters in serverside `SearchController` into methods which are shared between instant and full searches
 * Fix sorting (by sorting on the `pg_search_rank` column in Rails)
 * Rank everything onto a common scale for all searchable items (0..1 via tsearch normalization mode 32 + 8 + 2)
 * Add a nasty raw-sql hack to rank user results and maintain email searching without exposing too much info about email
   * If we're willing to drop email search (does anyone actually use it?) we can just replace this with just a pg_search on name
 * Rename `fuzzy_search_by_title` and `simple_search_by_title` to `instant_search` and `full_search`, respectively
 * Modify the `full_search` algorithm to just include trigram in the rankings initially
 * Add searching for Groups
 * Updated all indices to match the new search queries

Super Sekrit Internal Unsupported API™ Changes:

 * Change `type` parameter to separate `depth` and `scope` parameters.
   * Maintain `type=full` since it's 3 lines of code and it keeps Tenpenchii and Deadman working for now.  I know we don't officially support third party apps using the internal API, but users blame us anyways if they break so I'd rather not.  Plus Tenpenchii is off Google Play right now so updating might be slower (idk how this stuff works tbh)

Clientside Changes
 * Switch to the modified search endpoints and results
 * Add groups to dropdown
 * Move some of the stuff from the autocomplete to use `{{#link-to}}`
 * Use `transitionToRoute` to get to the `/search` page if you hit enter in the search box
 * Replace `var self = this` with `() => {}` functions where it was easy.

The clientside changes should reduce reboots of the app (searching no longer requires it to reboot) and the search changes should greatly improve result relevance.

I did some rough benchmarks on the search methods too, in development:
Cold Cache: 40ms instant, 600ms full
Hot Cache: 1.1ms instant, 1.5ms full

### DIFF
--- a/app/frontend/app/components/waifu-selector.js
+++ b/app/frontend/app/components/waifu-selector.js
@@ -24,12 +24,12 @@ export default Ember.Component.extend({
       },
       queryTokenizer: Bloodhound.tokenizers.whitespace,
       remote: {
-        url: '/search.json?query=%QUERY&type=character',
+        url: '/search.json?scope=character&depth=instant&query=%QUERY',
         filter: function(characters) {
           return Ember.$.map(characters.search, function(character) {
             return {
               value: character.name,
-              char_id: character.id
+              char_id: character.link
             };
           });
         }
--- a/app/api/api_v1.rb
+++ b/app/api/api_v1.rb
@@ -535,9 +535,9 @@ class API_v1 < Grape::API
   end
   get '/search/anime' do
     anime = Anime.accessible_by(current_ability).includes(:genres)
-    results = anime.simple_search_by_title(params[:query]).limit(20)
+    results = anime.instant_search(params[:query]).limit(20)
     if results.length == 0
-      results = anime.fuzzy_search_by_title(params[:query]).limit(20)
+      results = anime.full_search(params[:query]).limit(20)
     end
 
     title_language_preference = params[:title_language_preference]
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -16,10 +16,7 @@ class SearchController < ApplicationController
         query = params.require(:query)
 
         # The cheap price of supporting Tenpenchii and Deadman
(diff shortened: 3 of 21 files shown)


NuckChorris:
I just want to note that the results in some cases are somewhat... wrong (I think it might be related to the 8 or 2 normalizations).  So I'd like a dump of the `anime`, `manga`, and `groups` tables (cc @vikhyat) if possible, to tweak the ranking algorithm against :dango:

app/frontend/app/components/waifu-selector.js
```diff
@@ -24,12 +24,12 @@ export default Ember.Component.extend({
       },
       queryTokenizer: Bloodhound.tokenizers.whitespace,
       remote: {
-        url: '/search.json?query=%QUERY&type=character',
+        url: '/search.json?scope=character&depth=instant&query=%QUERY',
         filter: function(characters) {
           return Ember.$.map(characters.search, function(character) {
```
vikhyat: Weird that we are using jQuery `map` instead of `characters.search.map`.

vikhyat:
We should have a client-side SearchResult model or something, currently logic for executing search is distributed all over our routes. It doesn't need to inherit from `DS.Model`, `Ember.Object` would also be fine if we don't need ember-data functionality.

NuckChorris:
I actually mentioned that in the devchat, that we should create some central object to handle the search coordination.  I was leaning towards a service object which would wrap the bloodhound crap and expose the API as simple as setting options for depth/scope and then passing strings in to get a response.

I also was thinking about better options for the serialization/deserialization stuff, like AMS etc. but didn't find any good leads that direction.

Either way, I don't want the scope on this to creep too much, so I say we put those off.  Just calibrate the results on this and call it done.

vikhyat:
Cool, let's put it on the todo list then.

PR #419, opened 2015-02-20
Update auth routes, +Cleanup

vevix:
Gets rid of old devise controller overrides that we don't seem to use anymore, skips devise routes so we can use our new ember routes. :balloon: 

This will require updating the forums "Log in" link, I've created a PR for that here: hummingbird-me/discourse-hummingbird#4


vikhyat:
Awesome! Do we still redirect to the previous URL after sign in?

vevix:
Yeah we do, at least from my testing!

vikhyat:
Great!

PR #420, opened 2015-02-20
authenticate user when requesting news feed

vevix:
was there another way for a user to trigger this error without just manually submitting a JSON request?

```
NoMethodError (undefined method `id' for nil:NilClass):
  app/services/news_feed.rb:37:in `initialize'
  app/controllers/stories_controller.rb:14:in `new'
  app/controllers/stories_controller.rb:14:in `index'
  lib/log_before_timeout.rb:14:in `call'
```


NuckChorris:
iirc wasn't there also a place where `NewsFeed.new(nil)` happened in a worker?

NuckChorris:
:+1: LGTM, dunno what it fixes, but it's a line that should've been there from the start

PR #424, opened 2015-02-23
Group search fixes & tests

NuckChorris:
I'd meant to add thorough tests once I confirmed that the ugly presenter methods was the route we wanted to take, so here they are now.  This should confirm functionality for the most part.

Also fixed the leftover issue I missed when I blind-fixed the manga presenter as well as the image size adjustment for instant search (it was previously giving `:original` size for instant search which would result in excessive bandwidth usage)

Should also fix the unreported (for obvious reasons) issue with onboarding "element" search.  Maybe.  I'm not 100% certain of this due to unfamiliarity with the expected JSON response.  I don't think it will work, but I also don't know how to fix it.

Relevant is the JSON it produces: 

```json
{
    "search": [
        {
            "age_rating": "R17+", 
            "age_rating_guide": "Violence, Profanity", 
            ...
        }
    ]
}
```

Won't ED be confused by the `search` key?  I don't even know where it gets the `search` key from...

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -83,7 +83,7 @@ class SearchController < ApplicationController
   end
 
   # Presenters
-  def present_manga(manga, depth=:full)
+  def present_manga(manga)
     {
       type: 'manga',
       title: manga.canonical_title,
@@ -93,7 +93,7 @@ class SearchController < ApplicationController
       rank: manga.pg_search_rank,
       badges: [
         { class: 'manga', content: "Manga" },
-        { class: 'episodes', content: "#{x.volume_count || "?"}vol &bull; #{x.chapter_count || "?"}chap" }
+        { class: 'episodes', content: "#{manga.volume_count || "?"}vol &bull; #{manga.chapter_count || "?"}chap" }
       ]
     }
   end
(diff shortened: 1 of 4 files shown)


cybrox:
_LGTM_ from me from looking over the code two times.
I don't have a dev env with a sufficient dataset at hand to run the tests but I assume you've already done that anyways?

NuckChorris:
The tests run on the data in `fixtures/` with all of them passing, but my hand-testing dataset is sparse as fug

cybrox:
Uh, good point, I more thought about manually triggering a search to see if the output makes sense. But given the fact that it's mostly small format error fixes that cause nasty bugs, I don't think it's necessairy to dig into that any further anyways.

NuckChorris:
I did that for the most part, but as the OP says I had trouble with ActiveModelSerializers.  Trouble that can wait a little longer if nobody can spot an obvious cause.

cybrox:
If you mean the _weird_ `search` key you get on onboarding, that should not affect EmberData at all and iirc it's not stuffed into that key via ActiveModelSerializers.
The content of `search` is directly loaded into the ember store via `pushPayload` after the search results were returned.

NuckChorris:
Then I guess we're good.  I wasn't sure, it _seemed_ odd, but if it works it works.

cybrox:
It _IS_ odd, but I think it was the most convenient way to load search results into the `anime` store via ember data without actually using the `/anime` endpoint.

I wouldn't have noticed anything wrong in the changesm, so yea.

PR #425, opened 2015-02-23
Fix waifu selector dropdown titles

cybrox:
Not much to say about this. This fixes the waifu select drop down when a user is editing his profile details.

_waifu still broken apparently.  probably a typo went unnoticed in there_
~ @NuckChorris


NuckChorris:
rofl i'm an idiot

PR #426, opened 2015-02-24
First Notification refactoring

cybrox:
This pull request contains all the frontend changes for the basic new notification system. This includes:  
- A new notification page that displays all user notification (maximum now is 20)
 - Allows user to mark either one or all notification(s) as read.
- The notification tray has a bit less padding and will now display up to 6 notifications.

Image of Notification page

The new notification system handles notification messages with a message helper, the following _templates_ are currently implemented:
- feed_post (_old profile_comment_)
- feed_post_comment (_old comment_reply_)
- feed_post_reply
- feed_post_like
- user_follow

This will only add the needed frontend placeholders, no backend changes.
It will be deployed in a first step of the refactoring to get the new notifications page online and add the mark as read functionality. Any backend changes necessairy will be done later on.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -29,4 +29,14 @@ class NotificationsController < ApplicationController
       end
     end
   end
+
+  def mark_read
+    if params[:notification_id].nil?
+      Notification.where(user_id: current_user).update_all seen: true
+    else
+      Notification.where(id: params[:notification_id], user_id: current_user).update_all seen: true
+    end
+
+    render json: { success: true }
+  end
 end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -3,15 +3,15 @@ class NotificationsController < ApplicationController
 
   def index
     hide_cover_image
-    @notifications = Notification.where(user_id: current_user).order("created_at DESC").includes(:source, source: :target).limit(10)
+    @notifications = Notification.where(user_id: current_user).order("created_at DESC").includes(:source, source: :target).limit(20)
 
     respond_to do |format|
       format.json { render json: @notifications, each_serializer: NotificationSerializer }
       format.html { render_ember }
     end
     @notifications.where(seen: false).each {|x| x.update_column :seen, true }
-    if @notifications.count > 10
-      Notification.where(user_id: current_user, seen: true).order("created_at").limit(@notifications.count - 10).each {|x| x.destroy }
+    if @notifications.count > 20
+      Notification.where(user_id: current_user, seen: true).order("created_at").limit(@notifications.count - 20).each {|x| x.destroy }
     end
     Notification.uncache_notification_cache(current_user.id)
   end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -9,7 +9,6 @@ class NotificationsController < ApplicationController
       format.json { render json: @notifications, each_serializer: NotificationSerializer }
       format.html { render_ember }
     end
-    @notifications.where(seen: false).each {|x| x.update_column :seen, true }
     if @notifications.count > 20
       Notification.where(user_id: current_user, seen: true).order("created_at").limit(@notifications.count - 20).each {|x| x.destroy }
     end
--- a/app/frontend/app/helpers/notification-message.js
+++ b/app/frontend/app/helpers/notification-message.js
@@ -19,7 +19,7 @@ export function notificationMessage(notificationObject) {
 
   if(messageStore[messageType] === undefined) { return "A cat probably ate this notification!"; }
   var msg = messageStore[messageType].replace('<user>',
-    '<span class="who">'+notificationObject.get('source.user.username')+'</span>');
+    '<span class="who">'+notificationObject.get('sourceUser.username')+'</span>');
 
   return new Ember.Handlebars.SafeString(msg);
 }
--- a/app/frontend/app/helpers/notification-message.js
+++ b/app/frontend/app/helpers/notification-message.js
@@ -1,0 +1,27 @@
+import Ember from 'ember';
+
+export function notificationMessage(notificationObject) {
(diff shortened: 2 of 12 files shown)


app/frontend/app/helpers/notification-message.js
```diff
@@ -0,0 +1,27 @@
...
+export function notificationMessage(notificationObject) {
+  if(notificationObject === undefined) { return; }
+
+  var messageType = notificationObject.get('notificationType');
+  var messageStore = {
+
+    /* Old notification types! */
+    profile_comment: '<user> wrote a post on your feed.',
+    comment_reply: '<user> wrote a reply to your comment.',
+
```
NuckChorris: We can probably migrate these instead of providing a permanent fallback

app/frontend/app/templates/notifications.hbs
```diff
@@ -1,11 +1,35 @@
...
+    <div class="col-md-10 col-md-offset-1">
       <ul class="all-notifications">
-        {{#each}}
-          {{partial "notification"}}
+        {{#each notification in model}}
+          <div {{bind-attr class=":notification-item notification.seen::new"}}>
+            <div class="notification-title">
+              <a {{bind-attr href=notification.link}}>
+                <div class="avatar">
+                  {{avatar notification.sourceUser 'thumb_small'}}
```
NuckChorris: This rules out sourceless notifications doesn't it? :(

app/frontend/app/helpers/notification-message.js
```diff
@@ -0,0 +1,27 @@
...
+    feed_post:          '<user> wrote a post on your feed.',
+    feed_post_comment:  '<user> wrote a comment to your feed post.',
+    feed_post_reply:    '<user> commented on a feed post you commented on.',
+    feed_post_like:     '<user> liked your feed post.',
+    user_follow:        '<user> followed you.'
```
NuckChorris: It'd be nice to be able to delegate the whole hbs template for the row somehow. There's lots of more complicated notifications which don't come from a single user (likes coalesced, etc.) which would want logic on how many have been merged, etc. for example "5 users liked your post!" with stacked avatars or "Nuck liked your post!"

If it delegated to a specific hbs template (even if you just compile the templates and put them in a hash in here) based on type, all these more complex logics would be feasible.

Could it delegate to "notification-[type]" component? In future-ember there will be the `{{component}}` helper but right new theres minutebase/ember-dynamic-component

app/frontend/app/helpers/notification-message.js
cybrox: Indeed, I should probably remove that, I'm currently mostly using it for testing since this does not contain any backend changes yet.

app/frontend/app/templates/notifications.hbs
cybrox: Yes, as it currently is, all notifications are forced to have a source. We might need to change that if we want to bundle notifications. Unless we do something like _<source> and 3 others..._ to prevent bigger schema changes.

NuckChorris: I'm more thinking about cases where the source might be a show that just aired or your MAL import finishing

Stepepper:
No "<user> mentioned you in a post/comment." ?

NuckChorris:
Not yet, we need to do more before that

vevix:
@NuckChorris is right, but we do have a PivotalTracker story for that functionality: https://www.pivotaltracker.com/n/projects/1242706/stories/85563548

It may be a good idea to advertise the PT board in the README, considering its public.

Wopian:
And [here](http://forums.hummingbird.me/t/how-to-help-out-with-hummingbird/9390) when the forums isn't :skull:.

cybrox:
**So, what's the status of this**
- The new notifications page is added, tested and works as intended.
- Possibility to mark single/all notifications as read has been added.
- A new frontend helper has been added for adding simple template messages.
- User feed doesn't modify notification seen status anymore, only permalink does.

**Things that needs fixing**
- Right now, there is a strange bug that causes notifications to have the wrong user_id (the one who gets the notification, not the one who triggered it). I'll try to look into it but I have no idea what dumb mistake I made to cause that.
- Comment_reply Notifications do not yet get marked as read automatically, because I'm too noob to figure out what the best way to do it (performance wise) is. Requesting rails plebs help :fearful: 

Other than that, it's pretty much ready, just needs backend changes if we want to support "@" mentions but that's another thing.

app/controllers/notifications_controller.rb
```diff
@@ -30,4 +29,14 @@ def show
       end
     end
   end
+
+  def mark_read
```
NuckChorris: We should probably do this via REST instead of RPC, since that's the convention in both ED and Rails.  Ember-Data can set `seen: true` and then `.save()` the instance.  "Mark all read" could be done by issuing an update to a notification with magic id `all` — you might have to do a direct AJAX call but ED could handle the individual ones.

config/routes.rb
```diff
@@ -49,6 +49,7 @@
   get '/users/sign_in', to: redirect('/sign-in')
   get '/users/sign_up', to: redirect('/sign-up')
 
+  post '/notifications/mark_read(/:notification_id)' => 'notifications#mark_read'
   resources :notifications, only: [:index, :show]
```
NuckChorris: Again, if we use REST for this instead, it'll need the addition of `:update` to the `:notifications` resources and removal of the `/mark_read/` endpoint.

But if we do use RPC, it should also be `/notifications/:id/mark_read`, since `mark_read` is a method of the notification — not the other way around.

NuckChorris:
Needs rebase and I would definitely like to see it moved over to REST instead of RPC (it shouldn't be too hard to do, and it'd avoid adding another legacy RPC-style action that needs to be ED'd down the road) but otherwise looks like a damn solid first step towards better notifications.

NuckChorris:
So, now that 1.11 landed we also have the `component` helper for dynamic component name resolution which would be supah awesome for this

PR #427, opened 2015-02-24
Fix various bugs introduced by search refactoring

cybrox:
This will fix the follwoing things that **@NuckChorris broke™**

- Fix search in favorites adding wrong anime/manga.
- Quick fix for not using large images in fav edit preview.
- Not display already faved media in favorite box search
- Use medium images for search results (smaller images are broken)
 - Yes, this is a very hacky solution here, we should maybe finally sort out while images break in this case.
- Fix title of search results in the header instant search.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -50,7 +50,7 @@ class SearchController < ApplicationController
         end
 
         if depth == 'instant'
-          results.each { |x| x[:image] = x[:image].url(:small) }
+          results.each { |x| x[:image] = x[:image].url(:medium) }
         end
 
         return error! "No results", 404 if results.count == 0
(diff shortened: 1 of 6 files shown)


app/controllers/search_controller.rb
```diff
@@ -50,7 +50,7 @@ def search
         end
 
         if depth == 'instant'
-          results.each { |x| x[:image] = x[:image].url(:small) }
+          results.each { |x| x[:image] = x[:image].url(:medium) }
```
NuckChorris: This happened because not all types have `:small` sizes.  We should probably remedy this and standardize somewhat.

Until we get around to that, this'll work

NuckChorris:
lgtm :+1:

PR #428, opened 2015-02-26
Fix Facebook disconnection & Group stories

vevix:
Fixes Facebook disconnection expecting a POST but using a GET on the frontend, also fixes an issue where group stories weren't being added to the news feed of a group member if they didn't follow that user.

### DIFF
--- a/app/frontend/app/templates/settings.hbs
+++ b/app/frontend/app/templates/settings.hbs
@@ -148,7 +148,9 @@
         <div class="panel-heading"><h3 class="panel-title">Connections</h3></div>
         <div class="panel-body">
           {{#if currentUser.hasFacebook }}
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -111,7 +111,7 @@
     post :comment
     post :update_setting
 
-    post "/disconnect/facebook" => 'users#disconnect_facebook',
+    get "/disconnect/facebook" => 'users#disconnect_facebook',
```
NuckChorris: We should probably change this on the other side instead. GET should be idempotent

vevix: :+1: changes made :balloon:

NuckChorris:
LGTM :thumbsup:

PR #431, opened 2015-03-05
Make joining groups default to accepted

NuckChorris:
I've put in a `Group#public?` method on the grounds that eventually we'll want an actual enum behind it and allow for non-public groups (I guess probably private and invite-only)


vevix:
:+1: :shipit:

PR #432, opened 2015-03-05
serialize pending, and staff members first

vevix:
:arrow_up: :arrow_down:

### DIFF
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -4,11 +4,9 @@ class GroupMembersController < ApplicationController
   def index
     group = Group.find(params[:group_id])
 
-    userHasModPermissions = current_user &&
-      (group.has_admin?(current_user) || group.has_mod?(current_user))
-
-    members = userHasModPermissions ? group.members : group.members.accepted
-    members = members.page(params[:page]).per(20)
+    is_staff = current_user && group.is_staff?(current_user)
+    members = is_staff ? group.members : group.members.accepted
+    members = members.order(pending: :desc, rank: :desc).page(params[:page]).per(20)
     render json: members, meta: {cursor: 1 + (params[:page] || 1).to_i}
   end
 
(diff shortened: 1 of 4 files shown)


app/models/group.rb
```diff
@@ -108,8 +108,8 @@ def has_mod?(user)
     member(user).try(:mod?)
   end
 
-  def can_delete_story?(user)
-    has_admin?(user) || has_mod?(user)
+  def is_staff?(user)
+    !!(has_admin?(user) || has_mod?(user))
```
NuckChorris: Glad to see you renamed this, it was bothering me with how specific it was before! Also bang bang is the best operator :gun: :gun:

app/controllers/group_members_controller.rb
```diff
@@ -4,11 +4,9 @@ class GroupMembersController < ApplicationController
...
     group = Group.find(params[:group_id])
 
-    userHasModPermissions = current_user &&
-      (group.has_admin?(current_user) || group.has_mod?(current_user))
-
-    members = userHasModPermissions ? group.members : group.members.accepted
-    members = members.page(params[:page]).per(20)
+    is_staff = current_user && group.is_staff?(current_user)
+    members = is_staff ? group.members : group.members.accepted
+    members = members.order(pending: :desc, rank: :desc).page(params[:page]).per(20)
```
NuckChorris: This part looks solid to me :thumbsup:

app/frontend/app/routes/group/members.js
```diff
@@ -3,15 +3,6 @@ import Paginated from '../../mixins/paginated';
...
 export default Ember.Route.extend(Paginated, {
-  // use the recent members data we already have
-  model: function() {
-    // keep `cursor` at 1, as first visit we only have at most 15 members
-    // and pages are per 20.
-    return this.store.filter('group-member', (member) => {
-      return member.get('groupId') === this.modelFor('group').get('id');
-    });
-  },
-
```
NuckChorris: I have no idea what's going on in this bit, but otherwise this looks good :+1:

PR #433, opened 2015-03-06
Use MessageFormatter to format the user and group about text

NuckChorris:
Exactly as the title says, this uses the `MessageFormatter` service to expose formatting capabilities in the about section.  I haven't really tested it, because I'm just doing this between things today.  Merge with caution.

This should do [Convert newlines in "about" section to &lt;br&gt; tags](https://www.pivotaltracker.com/story/show/88584772)

### DIFF
--- a/app/frontend/app/models/group.js
+++ b/app/frontend/app/models/group.js
@@ -5,12 +5,14 @@ export default DS.Model.extend(ModelTruncatedDetails, {
   name: DS.attr('string'),
   bio: DS.attr('string'),
   about: DS.attr('string'),
+  aboutFormatted: DS.attr('string'),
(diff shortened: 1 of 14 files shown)


vevix:
Looks good :+1: Is there any worry about people putting YouTube/Gfy embeds in their about?

NuckChorris:
I for one encourage people to put more gifs and gfycats in their about box :D

JoshFabian:
Will this also use onebox when that is merged in?

NuckChorris:
Yep. This also provides spoilers (which probably won't work actually, we should fix that) and if we add markdown here it also adds it to feeds.

PR #435, opened 2015-03-09
Add a 'Your Groups' section to browsable groups page.

vevix:
:arrow_up: :arrow_down:

:balloon:

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -7,7 +7,7 @@ class GroupsController < ApplicationController
     groups = if params[:trending].present?
       Group.trending(page: page, per: limit)
     elsif params[:user_id].present?
-      User.find(params[:user_id]).groups.page(page).per(limit)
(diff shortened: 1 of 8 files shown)


app/controllers/groups_controller.rb
```diff
@@ -1,19 +1,20 @@
...
-    else # Get recent groups
-      Group.order('created_at DESC').take(6)
+    # query values
+    limit = (params[:limit] || 20).to_i
+    page = (params[:page] || 1).to_i
+
+    groups = if params[:trending].present?
+      Group.trending(page: page, per: limit)
+    elsif params[:user_id].present?
+      User.find(params[:user_id]).groups.page(page).per(limit)
```
NuckChorris: If `User#groups` doesn't have an order applied we should apply one here (`id ASC` would probably suffice) as [kaminari advises](https://github.com/amatsuda/kaminari/wiki/Kaminari-recipes#-dont-forget-to-add-an-order-scope)

NuckChorris:
LGTM :shipit:

PR #436, opened 2015-03-09
Fix winky

NuckChorris:
Fixes `;-)` as a group name.

By disallowing it.

Basically the Kim Jong Un approach to debugging.

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -90,7 +90,8 @@ class Group < ActiveRecord::Base
   validates :name,
     presence: true,
     uniqueness: {case_sensitive: false},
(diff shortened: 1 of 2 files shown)


vevix:
:+1: maybe a test would be a good addition

NuckChorris:
I prefer not to test simple validations because (1) I trust Rails tested them already and (2) they've got a tendency to change over time and confound future development. And (3) there's so little to be gained.

You're right that this would make a great regression test if I'd actually tried to figure out why the data is null. Perhaps I still will; I intend to dump a copy of the row before we delete it. I'll write a real test if/when I do that.

vevix:
:shipit: LGTM

PR #438, opened 2015-03-09
fix reply box disappearance on dashboard

vevix:
Pass user scope into the serializer, as this is executed by a sidekiq worker and `current_user` isn't in that context.


NuckChorris:
Naisu catch~ :shipit: LGTM

PR #441, opened 2015-03-09
Frontend Tests

vevix:
Adding a bunch of tests to our frontend :+1:

Very much a WIP.


NuckChorris:
Maybe it'd be good to just merge this anyways, even though it's WIP it adds tests and gets us started on that road?

vevix:
I fear these might need some changing with the ember update.

PR #442, opened 2015-03-10
Small tweaks

NuckChorris:
I did some tiny tweaks to the CSS on profile pages to make it not quite as shit on mobile, namely: scrolling the pillbar horizontally instead of wrapping it, and making it so the follow button wasn't so damn *huge* or *low*

Then I fixed the bug with `aboutDisplay` rendering the raw HTML.  I don't think this will be a security issue despite the fact that it theoretically exposes the raw about text without escaping, since users can no longer set the raw about text.  Assuming nobody previously crammed a `<script>` in there, we're probably solid.

And finally I added the group box to permalink pages in a group.

And then, post-finally, I added a new statistic in kotodama of "number of approved group joins in the past 24 hours" because why not

### DIFF
--- a/app/assets/stylesheets/partials/user-cover-image.css.scss
+++ b/app/assets/stylesheets/partials/user-cover-image.css.scss
@@ -289,6 +289,11 @@
         background: rgba(255,255,255,.3);
       }
     }
+    @media (max-width: 694px) {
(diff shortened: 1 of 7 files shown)


vevix:
:+1:

PR #444, opened 2015-03-10
add hummingbird onebox

vevix:
:arrow_up: :arrow_down:


NuckChorris:
LGTM, though I don't like the code style on the big HTML string.  A heredoc would be much better (and allow double quotes for the HTML!)

vevix:
Yeah we can change it up if needed, this was just a C/P from the forums.

PR #450, opened 2015-03-11
support strawpoll in onebox

vevix:
:100: :8ball: :shipit:


NuckChorris:
:facepunch: :v: :vhs: LGTM

PR #452, opened 2015-03-11
Bitcoin support

vevix:
:moneybag: :moneybag: :moneybag: :moneybag: :moneybag: :moneybag: :moneybag: :moneybag: :moneybag: :moneybag:

### DIFF
--- a/app/frontend/app/controllers/pro.js
+++ b/app/frontend/app/controllers/pro.js
@@ -35,9 +35,8 @@ export default Ember.ArrayController.extend(HasCurrentUser, {
   // Disable the buy button until stripe is loaded.
   disablePurchase: true,
   loadStripe: function() {
-    var self = this;
-    loadScript("https://checkout.stripe.com/v2/checkout.js").then(function() {
-      self.set('disablePurchase', false);
+    loadScript("https://checkout.stripe.com/v2/checkout.js").then(() => {
+      this.set('disablePurchase', false);
     });
   }.on('init'),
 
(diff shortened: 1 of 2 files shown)


app/frontend/app/controllers/pro.js
```diff
@@ -104,7 +103,9 @@ export default Ember.ArrayController.extend(HasCurrentUser, {
         key: PreloadStore.get('stripe_key'),
         name: "Hummingbird PRO",
         description: this.get('selectedPlan.name'),
-        token: tokenCallback
+        token: tokenCallback,
+        amount: parseInt(this.get('selectedPlan.amount') * 100),
```
NuckChorris: parseInt without a radix parameter of 10 isn't safe.  Specify it just to be safe.

Also shouldn't the multiplication be after it's turned into an integer, not before?

vevix: Nah because then it would parse `3.99` to `3`.

NuckChorris: Then `parseFloat`?

The parseInt is probably receiving a Number from the multiplication already and doing nothing except ensuring it's an integer (or converting Number -> String -> Number, I'm not sure how parseInt works internally).  If it's supposed to truncate, then `Math.floor` is probably more idiomatic?

vevix: Just tested and it's not needed at all. `this.get('selectedPlan.amount') * 100` works, without issue.

NuckChorris: :+1: Figured as much.  Dynamic types are magic like that :tophat: :sparkles:

NuckChorris:
:moneybag: :+1: :sparkles: :money_with_wings: :slot_machine:

PR #455, opened 2015-03-11
fix substory deletion issue

vevix:
:arrow_up: :arrow_down:

There was an issue with deleting substories > the first two. It would delete them, but not remove the object from display.


NuckChorris:
Lgtm

PR #458, opened 2015-03-12
Added site-feature-break-counter

cybrox:
We all know, we really need this. :bomb: :bomb:

### DIFF
--- a/app/frontend/app/routes/dashboard.js
+++ b/app/frontend/app/routes/dashboard.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import Paginated from '../mixins/paginated';
 import setTitle from '../utils/set-title';
 /* global MessageBus */
+/* global moment */
 
 export default Ember.Route.extend(Paginated, {
   preloadKey: "dashboard_timeline",
--- a/app/frontend/app/routes/dashboard.js
+++ b/app/frontend/app/routes/dashboard.js
@@ -33,7 +33,8 @@ export default Ember.Route.extend(Paginated, {
   setupController: function(controller, model) {
     controller.setProperties({
       'userInfo': this.store.find('userInfo', this.get('currentUser.id')),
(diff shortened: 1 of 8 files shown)


app/frontend/app/routes/dashboard.js
```diff
@@ -33,7 +33,8 @@ export default Ember.Route.extend(Paginated, {
   setupController: function(controller, model) {
     controller.setProperties({
       'userInfo': this.store.find('userInfo', this.get('currentUser.id')),
-      'trendingGroups': this.store.find('group', { trending: true, limit: 3 })
+      'trendingGroups': this.store.find('group', { trending: true, limit: 3 }),
+      'breakCounter': moment.unix(parseInt(window.genericPreload.break_counter)).fromNow().replace('ago', '')
```
NuckChorris: Won't genericPreload fail if the user comes to this page from another page on the site?

cybrox: Granted, it does, didn't think about that. Moment.js catches null-ish values though.
The thing is, I don't think this thing is worth an additional request, maybe we should just hide it if it's 0?

NuckChorris: Probably just hide if it's null.  I doubt many people will notice and if anyone asks we'll just say it doesn't count as a significant site feature :trollface:

cybrox: If anyone asks, we just say that the thing is apparently broken and that's so instant that it doesn't even show up :trollface:

NuckChorris:
:+1:

cybrox:
If this breaks anything in production, people will kill me :skull: :smile:

app/frontend/app/routes/dashboard.js
vevix: Probably would of been good to use our `PreloadStore` util and define a fetch, so that if the key doesn't exist we could return whatever.

PR #459, opened 2015-03-13
Adjustments to rails testing setup

NuckChorris:
Some tweaks to the testing system.  If all the tests pass with these, we're probably good.

So, some explanation regarding the changes:

#### Moving fakeweb fixtures to their own directory and switching to webmock
When we started with fakeweb there were only like two, now there's like 10 and it's cluttering up that directory.  Also I remember Rails saying somewhere that "the fixtures directory is for yaml fixtures"

Swiching to webmock allows us to test APIv1 (something we could also just use Rack::Test for) but also enables more powerful (and more importantly, more *localized*) testing stuff.  We can even assert on requests made!

#### Adding `json_expressions`
We need a DSL for specifying our responses as we move more and more towards Rails-as-an-API.  This isn't perfect, but there's really no great ones.  It'll do until we can test the serializers directly.

#### Adding `factory_girl`
Fixtures are great, and base data is nice to have.  But sometimes you need a bit more *oomph*, like in the groups tests.  I ran out of users to test all the unique combinations.  Factories solve that very real problem, and if we combine them with fixtures we can get the best of both worlds.

#### Adding `minitest-rails`
I'm tired of the shittiness of Test::Unit that I'm forced to use by Rails.  Spec-style assertions are a lot clearer and `let(:thing) {}` is hella boss.  This plugs MiniTest into Rails as a (hopefully) drop-in replacement.

#### Adding binstubs & `rails/spring`
Checked-in binstubs are highly encouraged in Rails 4, and spring is going to be a default in Rails 5.  They're also great together: boot times of `rake` and `rails` things are reduced heavily by installing spring (since it basically keeps the car running)


NuckChorris:
Thinking of replacing FakeWeb with WebMock too, because then we could mount the APIv1 rack app and ensure that it never changes (within parameters we can easily specify with json_expressions!)

vevix:
:clock11: :clock1130: :clock12: :fireworks: :fireworks: :fireworks: 

:+1: LGTM

NuckChorris:
:tanabata_tree: :mahjong: :100: Should I do the WebMock so we can test APIv1?

vevix:
Go for it :+1:

vevix:
:+1: Looking good LGTM

PR #461, opened 2015-03-14
Payment refactoring

NuckChorris:
Basic premise is that billing (PaymentMethod) is extracted from the stuff which juggles PRO (ProMembershipManager).  Right now there's places where it's hardwired to Stripe, but I've extracted it so that unwiring those will be a simple task.

I also removed the elements of the ProMembershipManager which updated billing info and moved that to the controller instead.

I also dropped the token column (and all the token-passing) because in both Stripe and Braintree, the token is a nonce.  We can't reuse it and it's not needed.  We only need the customer id.

### DIFF
--- a/test/factories/user.rb
+++ b/test/factories/user.rb
@@ -1,0 +1,15 @@
+FactoryGirl.define do
+  factory :user do
+    name { Faker::Internet.user_name }
+    email { Faker::Internet.email }
+    password { Faker::Internet.password }
+    avatar { Faker::Avatar.image }
+
+    trait :with_stripe do
+      billing_id { Stripe::Customer.create({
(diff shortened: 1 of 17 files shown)


vevix:
Looks good so far! :+1: :100:

test/factories/user.rb
```diff
@@ -0,0 +1,16 @@
+FactoryGirl.define do
```
NuckChorris: For some reason `annotate` didn't wanna annotate the factories. It's supposed to Just Work™ but it didn't.

vevix: Probably didn't work because IIRC it expects `users.rb` for a Factory

PR #462, opened 2015-03-14
Fix comment and about section embeds for images and giphycat

JoshFabian:
Let me know what other embeds are screwing up in comments / about section. Please add to pivotal!

### DIFF
--- a/app/assets/stylesheets/base/layout.css.scss
+++ b/app/assets/stylesheets/base/layout.css.scss
@@ -27,6 +27,10 @@ html { overflow-y: auto; }
   &:hover { i { color: white; }}
 }
 
+img, iframe {
+  max-width: 100%;
+}
+
 
 .navbar-brand {
   background-image: url("/assets/logo-sprite.png");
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/base/layout.css.scss
```diff
@@ -27,6 +27,10 @@ html { overflow-y: auto; }
   &:hover { i { color: white; }}
 }
 
+img, iframe {
+  max-width: 100%;
+}
+
```
NuckChorris: This won't fuck anything else up will it? I can't imagine it would but side effects are real

NuckChorris:
Legtom :dango: :10: :frog: :tulip:

app/assets/stylesheets/base/layout.css.scss
JoshFabian: It shouldn't, we're basically say and image or iframe _isn't_ allowed to be larger than it's parent container.

PR #465, opened 2015-03-15
Fix cover image uploads, remove ridiculous production check

vevix:
:arrow_up: :arrow_down:

Recent Onebox CSS update applied `max-width: 100%` to all `` elements, which fucked with the way we do cropping.


NuckChorris:
Lgtm

PR #466, opened 2015-03-16
Fix removing single story liker

vevix:
We have a :beetle: at the moment, where if you are the only _liker_ on a story, and you unlike that story, your avatar doesn't actually get removed.

This fixes that. :beetle: :gun:


NuckChorris:
Lgtm

PR #469, opened 2015-03-18
Trending groups via SQL view

NuckChorris:
>[Active Record] was built on the notion that SQL is neither dirty nor bad, just verbose in the trivial cases. The focus is on removing the need to deal with the verbosity in those trivial cases (writing a 10-attribute insert by hand will leave any programmer tired) but keeping the expressiveness around for the hard queries—the type SQL was created to deal with elegantly.

— DHH, [Agile Web Development with Rails](http://event2011.googlecode.com/files/Agile_Web.pdf)

This basically calculates the z-score via two SQL views, one of which is a "materialized view".  One view is used to generate a "histogram" of GroupMembers by created_at.  The other (materialized) view then uses this data to calculate the z-score.

The materialized view is updated every 5 minutes by a Sidetiq worker.

On display, we remove groups the current user is in already.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -5,7 +5,7 @@ class GroupsController < ApplicationController
     page = (params[:page] || 1).to_i
 
     groups = if params[:trending].present?
-      Group.trending(page: page, per: limit)
+      Group.trending(current_user).page(page).per(limit).groups
     elsif params[:user_id].present?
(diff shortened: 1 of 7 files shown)


vevix:
Looks good, maybe @cybrox can check SQL as well -- not my strong point. :+1: from me.

cybrox:
Looks like sql madness to me.
Logic-wise, I don't see anything wrong, so I'll drop an LGTM :dromedary_camel:

PR #470, opened 2015-03-18
Announce deploys into #notifications

NuckChorris:
A halfassed but functional solution to tell us when things are deploying

I already deployed SLACK_WEBHOOK_URL and configured the integration so this is g2g


vevix:
:+1:

PR #471, opened 2015-03-19
update onebox, add gfycat to protocol rewrite list

vevix:
Updates onebox so that it has gfycat support by default, and an update to github embeds. Also makes it so gfycat embeds use `//` instead of `http://`.

:arrow_up: :100: :sparkle: :sparkles:


NuckChorris:
Lgtm

PR #474, opened 2015-03-21
allow admins to flag stories as NSFW

vevix:
:arrow_up: :arrow_down: :question: :checkered_flag:


NuckChorris:
Hah that works, lgtm

PR #476, opened 2015-03-22
Added googleadservice tracking to landing page

cybrox:
:boar: Just loaded the script globally for now since it's apparently bould to some pageloads events, we might want to optimize that later on. - Or just leave it if we're gonna track other buttons (like the normal header) as well.

### DIFF
--- a/frontend/app/controllers/index.js
+++ b/frontend/app/controllers/index.js
@@ -18,5 +18,37 @@ export default Ember.Controller.extend({
 
   hideOnMobile: function() {
     return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);
-  }.property()
+  }.property(),
+
+  googReportConversion: function(url) {
+    window.google_conversion_format = "3";
+    window.google_is_call = true;
+    var opt = {
+      onload_callback: function() {
+        if (typeof(url) !== 'undefined') {
+          window.location = url;
+        }
+      }
+    };
+    var conv_handler = window['google_trackConversion'];
+    if (typeof(conv_handler) === 'function') {
+      conv_handler(opt);
+    }
+  },
+
+  actions: {
+    emailSignupClick: function() {
+      window.google_conversion_id = 992672164;
+      window.google_conversion_label = "FrnFCPSZy1oQpPOr2QM";
+      window.google_remarketing_only = false;
+      this.googReportConversion("/sign-up");
+    },
+
+    facebookSignupClick: function() {
+      window.google_conversion_id = 992672164;
+      window.google_conversion_label = "s9hgCKiYy1oQpPOr2QM";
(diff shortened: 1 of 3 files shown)


frontend/app/controllers/index.js
```diff
@@ -18,5 +18,37 @@ export default Ember.Controller.extend({
...
-  }.property()
+  }.property(),
+
+  googReportConversion: function(url) {
+    window.google_conversion_format = "3";
+    window.google_is_call = true;
+    var opt = {
+      onload_callback: function() {
+        if (typeof(url) !== 'undefined') {
+          window.location = url;
```
NuckChorris: Can't we do `transitionTo`?

NuckChorris:
Maybe `ember-cli-Google-analytics` would be better?

JoshFabian:
Are you certain they're the same thing? This embed code came from google adwords dashboard, not google analytics.

cybrox:
I only searched for google adservice, fml.
Does `ember-cli-google-analytics` support what we need?

NuckChorris:
Nah, is there an Adwords tracking plugin for ember-cli?

Anyways the biggest things that bug me in this are that (1) it's just kind of crammed in the rails template (2) if Google fails, God forbid, it just stops transitioning to the sign up pages and (3) it's not using `transitionTo` which causes a reload right as we want to impress the user with our snappiness

cybrox:
(1) As I said, I just added it globally, we need to find a better way for this. (2) Is a valid point, maybe a timeout would be a workaround. And (3) Would be possible to change with a bit of hacking, though the Facebook links were never ember'd iirc.

NuckChorris:
The Facebook links actually can't be embered I think. May I suggest we just have a function that fire-and-forgets to GA? Then we'd just in the action put the `transitionTo`

cybrox:
That would be the only alternative we'd have if we want to drop using the callback for redirecting - which would definitely be a good idea.

vevix:
Don't we already have Segment, don't they have support for this?

JoshFabian:
Segment is regular google analytics but this is google adwords tracking.

NuckChorris:
This isn't it? https://segment.com/docs/integrations/adwords/

JoshFabian:
Oh, actually it is! @cybrox can we track button clicks with that? Just page tracking won't work otherwise we'll have 100% conversions since everyone will be hitting the landing page. Alternatively, if we fix Facebook to properly push into onboarding we could track the first page of onboarding, we just won't know how many people are FB connecting and how many are signing up with email. These are metrics I'd like to know because it could drive what we should be placing priority on in the future.

Either way, would be great to get tracking up in the next day or so. I'd like to start pushing traffic to this page.

vevix:
This seems to have stagnated and is pretty important. I can take this over and implement segment if @cybrox is too busy.

PR #479, opened 2015-03-26
Add prerendering hooks

NuckChorris:
Hopefully someday Ember FastBoot™ will be a reality but we sadly live in the world of today, and so we can't do `app.visit('/anime/sword-art-online').then()` and have to resort to this hack for now.

Hack taken from https://github.com/zipfworks/ember-prerender and modified.  Intent is to use ember-prerender since it's a fork of prerender.io with better performance properties and deterministic readiness checking via [XContentReady](https://github.com/n-fuse/the-XContentReady-Event/).

Workaround for emberjs/ember.js#9938 because FastBoot™ is gonna be a while

### DIFF
--- a/frontend/app/app.js
+++ b/frontend/app/app.js
@@ -15,10 +15,47 @@ if (!window.location.href.match('/sign-in')) {
   window.lastVisitedURL = window.location.href;
 }
 
+if (document.createEvent) {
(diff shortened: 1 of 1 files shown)


vevix:
Looks good :+1:

PR #489, opened 2015-04-04
Add MAL id to APIv1 anime responses

BeholdMyGlory:
This change adds the MAL id to the hash returned by `present_anime` in `api_v1.rb`.

There are many services today that integrate with MAL. In many cases this is likely done by keeping some kind of mapping between the MAL id and the local id of anime entries, much like Hummingbird does. Including the MAL id when using Hummingbird's API would make it easier to extend these services to integrate with Hummingbird.

### DIFF
--- a/app/api/api_v1.rb
+++ b/app/api/api_v1.rb
@@ -91,6 +91,7 @@ class API_v1 < Grape::API
       if anime
         json = {
           id: anime.id,
+          mal_id: anime.mal_id,
           slug: anime.slug,
           status: anime.status,
           url: "https://hummingbird.me/anime/#{anime.slug}",


erengy:
Note that API v1 has an undocumented `include_mal_id` [parameter](https://github.com/hummingbird-me/hummingbird/blob/be2ecf28b7d21ff36b2be43270be1982b416aace/app/api/api_v1.rb#L379), although it's limited to `/libraries` endpoint.

In addition, API v2 supports lookup by [MAL ID](https://github.com/hummingbird-me/hummingbird/blob/be2ecf28b7d21ff36b2be43270be1982b416aace/app/controllers/api/v2/anime_controller.rb#L17) (e.g. `myanimelist:1`). As far as I'm aware, you can make bulk requests as well (e.g. `1,2,3,4,5`).

BeholdMyGlory:
Ah, I missed that part. I guess with this change the MAL id is duplicated: once in the root of the hash, and once in the "anime" subhash. Considering that `include_mal_id` is undocumented, and doesn't seem to be used anywhere else in the codebase, simply removing the parameter and relying on `present_watchlist` (and in turn `present_anime`) to include the MAL id seems like the better option, as far as consistency is concerned (as `mal_id` cannot be included in the root of the hash in cases where the response includes more than one anime). Is there a reason in particular that you'd prefer to keep it optional? Given that it's JSON we're talking about, the addition of a single key/value pair shouldn't affect clients as far as I can tell.

Being able to request a single anime using the MAL id with API v2 might be useful in some instances, but would not be very helpful in the case where you want to map e.g. a Hummingbird user's "currently watching" list onto MAL ids.

NuckChorris:
We used to have this but I believe it was consciously removed because, quite simply, we don't want to expose API users to the horrors of MAL-to-Hummingbird mappings.  Some shows don't have equivalents, and our databases will not forever line up.

Like, we want to make it easy to switch things to use Hummingbird but it's kind of a double-edged blade if it enables people to continue using MAL IDs internally.

BeholdMyGlory:
Well, as long as a service supports both MAL and Hummingbird it's going to need some kind of mapping between the two, and pragmatically speaking, as long as Hummingbird already knows that mapping it would be better if it shared this information rather than having every service reimplement the same mapping. In the case where a Hummingbird entry doesn't have a MAL equivalent, simply returning null for the MAL id would suffice in my opinion (with a warning regarding this in the documentation, of course), serving as a signal that the entry in question needs to be specially handled.

Epictek:
@BeholdMyGlory I've been wanting to make a service to map multiple sites for a while now, Have not made any progress on it yet but I might try to get hummingbird to mal mappings implemented tomorrow or sometime later this week.

https://github.com/Epictek/AnimeID (no commits yet)

erengy:
I think a number of people already have such an ID map, for the vast majority of the database. We could use those as a seed, or get the most recent one from Hummingbird. However, there are more troublesome things to do:

1. The list must be kept up to date. It's possible, but there's no clean way to automate this process.
2. If we were to do this, it'd be more meaningful to include other popular services (AniDB, AniList, Anime News Network, Anime-Planet) as well, not just Hummingbird and MyAnimeList.
3. Not all entries line up one-for-one. What to do when an ID corresponds to multiple IDs at another service? How to handle similar inconsistencies?

erengy:
I'd like to add that keeping the database up-to-date is not limited to adding new entries. Taking MAL as an example, recent entries are sometimes invalidated due to various reasons. Some of them end up being merged. When their database guidelines get updated, relevant changes go back to older entries as well. For MAL, one could probably simplify the process by creating a new user account, adding every anime to its list, and then tracking that list via malappinfo.php. But that's just MAL; I don't know if other services could benefit from a similar approach.

NuckChorris:
I dunno if that actually even matters. I'm pretty sure we never change the MAL IDs and we seem to be fine.  It's probably such a small problem that you can fix them as they come up.  I'm assuming that's how all the current individual mapping systems work.

JoshFabian:
> Like, we want to make it easy to switch things to use Hummingbird but it's kind of a double-edged blade if it enables people to continue using MAL IDs internally.

I don't think this is a position we can take while 90% of the industry is still using MAL. It's not realistic to expect developers working with MAL currently to drop it and ask all of their users to move to HB. A bunch of small wins is how we continue to grow. When we're the industry leader, we can start dictating things like that, but forcing non-compatibility with MAL now seems very preemptive.

NuckChorris:
> It's not realistic to expect developers working with MAL currently to drop it and ask all of their users to move to HB. A bunch of small wins is how we continue to grow. When we're the industry leader, we can start dictating things like that, but forcing non-compatibility with MAL now seems very preemptive.

I'm mostly worried that we're gonna have all hell pulling that rug out from under developers when it comes time.  At the very least I want to put a big red flag in the docs advising developers not to key on MAL ID.

For now though, I'll give in and merge this :+1:

Epictek:
:+1:

BeholdMyGlory:
@NuckChorris Before merging, though, we need to decide what to do about the `include_mal_id` parameter and the related duplicate `mal_id`. Are you OK with my idea to just remove `include_mal_id` since it's undocumented?

erengy:
I still lean toward making it optional via `include_mal_id` parameter, if it's not going to complicate things in the codebase. Otherwise we end up either having duplicate `mal_id`s or breaking backward compatibility.

> Are you OK with my idea to just remove `include_mal_id` since it's undocumented?

Please don't remove anything from API v1, even if it's undocumented. It might break existing applications.

NuckChorris:
As much as I want to say "yes, please remove it", APIv1 is always backwards-compatible which means we can't even remove our stupid mistakes.  So I guess it's here to stay.

@erengy I don't see a reason to make it optional, it's a single field.  include_genres is similarly strange.  And anyways I give exactly zero shits about APIv1's code quality nowadays, it's a wretched section of the codebase and the sooner we get better APIv2 coverage, the happier I'll be.

BeholdMyGlory:
I personally don't think there would be any point to making it optional, as the only instance where the entry would be duplicated is in `/libraries/{id}`, and even that is only when the user explicitly issues a request using an undocument parameter---not a big deal, in other words.

Additionally, the behaviour of an `include_mal_id` parameter would be inconsistent: with `/libraries/{id}` (or any request that only returns a single anime entry?) the MAL id would be included in the root of the returned hash, while for other responses it would included in the anime subhash(es).

erengy:
Now that I think about it, we can't really avoid having duplicate `mal_id`s even if we make it optional, which was my entire concern.

All right, then. :+1:

NuckChorris:
Aight mergening

vevix:
Just an update, this is now live.

PR #492, opened 2015-04-11
add CustomEvent polyfill

vevix:
Fixes the issue with IE and the prerender code.


NuckChorris:
I just realized there were places I forgot to condition on `if (CustomEvent)`. But this works too so :+1:

PR #493, opened 2015-04-13
fix pro renewal worker

vevix:
:blush: :unamused:


NuckChorris:
:+1: how long has this been broken and how haven't we noticed? And how didn't my tests catch it

vevix:
Day 1 :tada: :cake:

PR #494, opened 2015-04-13
convert to symbol in MyAnimeListScrapeWorker

vevix:
This will fix a lot of our failed Sidekiq workers


NuckChorris:
Wow where do we even screw this one up but otherwise :+1: though I guess we should fix whatever is making the issues

vevix:
Yeah, we do pass it as a symbol but I assume Sidekiq doesn't care.

PR #495, opened 2015-04-16
remove dropbox info if unauthorized

vevix:
Will fix a lot of our dead workers when users remove the Hummingbird app from their Dropbox account.


NuckChorris:
Looks solid to me. This doesn't catch things like the out of space errors, right? I want to make those show an actual error to the user

vevix:
I believe that throws its own error!

PR #500, opened 2015-04-21
Rails controller cleanup 1

NuckChorris:
Need to do some poking around and final testing but everything seems to be fine.  And almost 1100 lines lighter.  Fuck yeah.

Two big parts to this: (1) tweaks to the helpers for rendering ember stuff and (2) deletion of old unused Rails views


app/controllers/application_controller.rb
```diff
@@ -57,18 +57,49 @@ def preload_objects
...
+    else
+      render json: {error: message}, status: status
+    end
   end
 
-  def hide_cover_image
-    @hide_cover_image = true
-  end
+  # DEPRECATED
+  def hide_cover_image;end
```
NuckChorris: Note to self: grep (or `ag` more likely) for all the `hide_cover_image` calls and rip them out

NuckChorris:
Just wanna add that this may fix `WARN: tilt autoloading 'coffee_script' in a non thread-safe way; explicit require 'coffee_script' suggested.` in the test suite — there were inline coffeescript blocks in some views.  If there are any remaining we can bake them into JS in a later PR.

vevix:
:tada: Cleanup :tada:

Issue #491, opened 2015-04-07
Fix MAL synopsis parser, clean existing synopses


erengy:
Example: https://hummingbird.me/anime/teekyuu-5

```
Fifth season of Teekyuu series.
googletag.cmd.push(function() {
googletag.defineSlot('/84947469/02PCAnimeSAFEtopLB01', [728, 90], 'div-gpt-ad-1425259000548-0').addService(googletag.pubads());
googletag.pubads().enableSingleRequest();
googletag.enableServices();
});




googletag.cmd.push(function() { googletag.display('div-gpt-ad-1425259000548-0'); });



```

MAL: http://myanimelist.net/anime/30567/Teekyuu_5

```html
<td valign="top"><h2>Synopsis</h2>Fifth season of <em>Teekyuu</em> series.<div class="border_top" style="padding:16px 0px 0px 0px;margin:14px 0px 0px 0px;"><div style="width:728px; height:90px; margin:0 auto"><script type='text/javascript'>
googletag.cmd.push(function() {
googletag.defineSlot('/84947469/02PCAnimeSAFEtopLB01', [728, 90], 'div-gpt-ad-1425259000548-0').addService(googletag.pubads());
googletag.pubads().enableSingleRequest();
googletag.enableServices();
});
</script>
<!-- 02PC Anime - SAFE - top - LB01 -->
<div id='div-gpt-ad-1425259000548-0' style='width:728px; height:90px;'>
<script type='text/javascript'>
googletag.cmd.push(function() { googletag.display('div-gpt-ad-1425259000548-0'); });
</script>
</div>
</div></div></td>
```

Could be a common issue. Probably because [this line](https://github.com/hummingbird-me/hummingbird/blob/057e522d03a0253650500c82745c5137b875d3b0/lib/mal_import.rb#L105) doesn't ignore text within `script` elements.

NuckChorris:
Fixed in #503

PR #503, opened 2015-04-22
Fixed issue #491 - Fix MAL synopsis parser, clean existing synopses

v4lproik:


NuckChorris:
Hacky as hell but should work.  And since Nokogiri doesn't have any better solution (fuck 'em for not implementing `text` like `innerText`) :thumbsup:

vevix:
Thanks :+1:

PR #511, opened 2015-04-26
fix :beetle:'s with current-user service

vevix:
Fixes issues with new current user injection and saving data, also cleans up the initializer.

### DIFF
--- a/frontend/app/initializers/current-user.js
+++ b/frontend/app/initializers/current-user.js
@@ -1,19 +1,13 @@
-import currentUserProxy from '../services/current-user.js';
-
 export function initialize(container, app) {
-  var user, store = container.lookup('store:main');
-
-  app.register('service:current-user', currentUserProxy);
+  ['route', 'view', 'component', 'controller'].forEach((component) => {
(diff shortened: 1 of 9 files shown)


frontend/app/initializers/current-user.js
```diff
@@ -1,19 +1,13 @@
-import currentUserProxy from '../services/current-user.js';
-
 export function initialize(container, app) {
-  var user, store = container.lookup('store:main');
-
-  app.register('service:current-user', currentUserProxy);
```
NuckChorris: Don't we still need to register it into the container?

vevix: ember-cli will handle that

NuckChorris: I thought that was a later addition in ember-cli, around Ember 1.10?

vevix: They made it so you can inject services into objects around that version with syntax like:
`currentUser: Ember.service.inject()`

but this is just ember-resolver magic.

NuckChorris:
Sweet, LGTM

PR #514, opened 2015-04-26
Fix Erroneous Ellipsis Overflow

wopian:
I believe this was designed for the "show more" cut-off point - but this doesn't work there anyway... :confounded: 

This solves a somewhat rare issue in Blink's/Chrome(?) rendering which will add an ellipsis for any lines that "overflow" the container, instead of dropping the word to the next line.

Example:
 
 On the last full line the word "subs" is right up against the container's width and Chrome changes it to "su...".

Without `text-overflow: ellipsis`:

The text is **not** overflowing the container.

### DIFF
--- a/app/assets/stylesheets/partials/status-panel.css.scss
+++ b/app/assets/stylesheets/partials/status-panel.css.scss
@@ -172,7 +172,6 @@
     padding: 0;
     margin-bottom: 0;
     overflow: hidden;
-    text-overflow: ellipsis;
     max-height: 500px;
 
     img {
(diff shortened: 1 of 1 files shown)


NuckChorris:
This might actually be for long links — to prevent them from overflowing the box?

wopian:
>This might actually be for long links — to prevent them from overflowing the box?

Why did I not think of this :facepunch: :dizzy_face:

wopian:
Actually a solution:

---

Remove `text-overflow: ellipsis` as done here,

Add `text-overflow: ellipsis` and `display: block` to `.comment-text a`

vevix:
:pray: update PR!

PR #516, opened 2015-04-28
redirect facebook login to onboarding

vevix:
Only assuming this works, which it should. Can't test Facebook locally without jumping thru hoops :(.


NuckChorris:
MGMT :+1:

PR #519, opened 2015-04-29
Hotfix for Genre Padding

wopian:
Changed Base `inline-css` to `padding-right: 0` as the only element importing `inline-css` and **using** the default 10px was onboarding. Everything else (or attempted to) overwrote it.

Fixes PR 422.


wopian:
Seems like I forgot to check the anime/manga pages for issues. #529 

Studios and genres no longer have padding.

PR #522, opened 2015-05-04
show feedback on simple onboarding

vevix:
When using the `simple` rating system, it doesn't show any feedback about which rating you have chosen on the onboarding section.

Also, the code for onboarding was referencing `media.libraryEntry` which failed on manga and made it always try to create a new library entry which would fail on the backend.


NuckChorris:
LGTM :shipit:

PR #523, opened 2015-05-04
Bump search results per-type to 7

NuckChorris:
Clearly 3 isn't enough according to http://forums.hummingbird.me/t/anime-entries-missing-in-search/22292


vevix:
:+1:

PR #524, opened 2015-05-04
Revert canonicalize URL changes, pending a better solution

NuckChorris:


vevix:
Can you also set the manga library route to #ember? :+1:

PR #526, opened 2015-05-04
Add paranoia gem

NuckChorris:
This gem allows us to _fake_ deleting things.  Meaning, it's not actually deleted, it just pretends to be.  So we have an audit trail.

For example, if a user is being harassed and they delete the posts, we'd probably still want to see the posts.  Right now, the only way to view them is to go into rails_admin (I assume it'll work in there, not sure though actually) but in future I'd like to serialize `deleted_at` and display deleted posts (with visual distinction obviously) to administrators


vevix:
Are we going to need to add anything to our queries to not serialize deleted posts?

NuckChorris:
Nope, this gem overrides `default_scope` on the model.  I even checked that `find` queries will return nothing

vevix:
Okay awesome :+1:

NuckChorris:
Rebased, ready for merge?

vevix:
:+1:

PR #530, opened 2015-05-05
Fix #512: posterImageThumb returning non-existent Images

### LINKED ISSUE #512: posterImageThumb returning non-existent Images
Medium and original sized images keep the original extension. Large is converted to jpg.
Most poster images work fine, because most were originally jpg in the first place.
posterImageThumb replaces /large/ with /medium/, though, so .png's would return 404, for instance.

https://static.hummingbird.me/anime/poster_images/000/003/936/large/gqt6iYB.jpg - JPEG
https://static.hummingbird.me/anime/poster_images/000/003/936/medium/gqt6iYB.png - PNG
https://static.hummingbird.me/anime/poster_images/000/003/936/medium/gqt6iYB.jpg - Non-existent

posterImageThumb is used by the favorites editor.
It's defined in /frontend/app/models/anime.js

DNA:
Hello guys! 

This is my first Pull Request ever, so let me know if I missed something important :)

This PR has 2 commits intended to fix an error reported on issue #512, where images wasn't showing properly when you click to edit the favorite list. This code solves the problem, but I think I can work a little further, like:

- The method `poster_image_thumb` name holds little meaning now that it have two uses. Maybe refactor this to `image_path` or something like that?
- Move the path generation from Ember to Rails and return `medium_image` instead of `original_image`?
- Maybe move this attachments and paths into a concern, as this is used on two models, so we can DRY it a little?
- There wasn't any tests regarding image paths, but I don't know if it's necessary and/or how to start this. 

What you guys think? :)

### DIFF
--- a/app/serializers/anime_serializer.rb
+++ b/app/serializers/anime_serializer.rb
@@ -2,9 +2,9 @@ class AnimeSerializer < ActiveModel::Serializer
   embed :ids
 
   attributes :id, :canonical_title, :english_title, :romaji_title, :synopsis,
-    :poster_image, :show_type, :age_rating, :age_rating_guide, :episode_count,
-    :episode_length, :started_airing, :started_airing_date_known,
-    :finished_airing, :genres, :updated_at
+    :poster_image, :poster_image_thumb, :show_type, :age_rating,
+    :age_rating_guide, :episode_count, :episode_length, :started_airing,
+    :started_airing_date_known, :finished_airing, :genres, :updated_at
 
   def id
     object.slug
@@ -23,6 +23,14 @@ class AnimeSerializer < ActiveModel::Serializer
       "/assets/missing-anime-cover.jpg"
     else
       object.poster_image.url(:large)
+    end
+  end
+
+  def poster_image_thumb
+    if !object.sfw? && (scope.nil? || scope.try(:sfw_filter))
+      "/assets/missing-anime-cover.jpg"
+    else
+      object.poster_image.url(:medium)
     end
   end
 
--- a/app/serializers/manga_serializer.rb
+++ b/app/serializers/manga_serializer.rb
@@ -5,6 +5,7 @@ class MangaSerializer < ActiveModel::Serializer
              :romaji_title,
              :english_title,
              :poster_image,
+             :poster_image_thumb,
              :synopsis,
              :chapter_count,
              :volume_count,
@@ -23,6 +24,10 @@ class MangaSerializer < ActiveModel::Serializer
   end
 
   def poster_image
+    object.poster_image.url(:large)
+  end
+
+  def poster_image_thumb
     object.poster_image.url(:large)
   end
 
--- a/frontend/app/models/anime.js
+++ b/frontend/app/models/anime.js
@@ -10,6 +10,7 @@ export default Media.extend(ModelCurrentUser, {
   romajiTitle: DS.attr('string'),
   synopsis: DS.attr('string'),
   posterImage: DS.attr('string'),
+  posterImageThumb: DS.attr('string'),
   showType: DS.attr('string'),
   ageRating: DS.attr('string'),
   ageRatingGuide: DS.attr('string'),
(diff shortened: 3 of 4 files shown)


app/models/anime.rb
```diff
@@ -83,16 +83,16 @@ class Anime < ActiveRecord::Base
...
-  def poster_image_thumb
+  def poster_image_thumb(size = :large)
     if self.poster_image_file_name.nil?
       "https://hummingbird.me/assets/missing-anime-cover.jpg"
     else
       # This disgusting fastpath brought to you by the following issue:
       # https://github.com/thoughtbot/paperclip/issues/909
       if Rails.env.production?
-        "https://static.hummingbird.me/anime/poster_images/#{"%03d" % (self.id/1000000 % 1000)}/#{"%03d" % (self.id/1000 % 1000)}/#{"%03d" % (self.id % 1000)}/large/#{URI.escape(self.poster_image_file_name.rpartition('.')[0])}.jpg?#{self.poster_image_updated_at.to_i}"
+        "https://static.hummingbird.me/anime/poster_images/#{"%03d" % (self.id/1000000 % 1000)}/#{"%03d" % (self.id/1000 % 1000)}/#{"%03d" % (self.id % 1000)}/#{size.to_s}/#{URI.escape(self.poster_image_file_name.rpartition('.')[0])}.jpg?#{self.poster_image_updated_at.to_i}"
```
fractalemagic: This fix won't work in production? The file extension is hardcoded to `jpg` here.

app/models/manga.rb
```diff
@@ -86,16 +86,16 @@ def canonical_title
...
-  def poster_image_thumb
+  def poster_image_thumb(size = :large)
     if self.poster_image_file_name.nil?
       "https://hummingbird.me/assets/missing-anime-cover.jpg"
     else
       # This disgusting fastpath brought to you by the following issue:
       # https://github.com/thoughtbot/paperclip/issues/909
       if Rails.env.production?
-        "https://static.hummingbird.me/manga/poster_images/#{"%03d" % (self.id/1000000 % 1000)}/#{"%03d" % (self.id/1000 % 1000)}/#{"%03d" % (self.id % 1000)}/large/#{self.poster_image_file_name}?#{self.poster_image_updated_at.to_i}"
+        "https://static.hummingbird.me/manga/poster_images/#{"%03d" % (self.id/1000000 % 1000)}/#{"%03d" % (self.id/1000 % 1000)}/#{"%03d" % (self.id % 1000)}/#{size.to_s}/#{self.poster_image_file_name}?#{self.poster_image_updated_at.to_i}"
```
vevix: File name should be escaped, and no idea why the poster image `large` style doesn't match anime. /cc @NuckChorris

vevix:
I'd like to see us actually use `Rails.cache.fetch` since we write the `:large` style to cache in our `after_save` hook.

I'm thinking serializing a template similar to what we do on the forums might be the best way. We can then create computed properties on the frontend model for each style that we need.

DNA:
You were right, anime image extension was hardcoded and manga image wasn't escaped from the start, this should be more consistent now :)

And I saw the cache saving, but didn't touch it since I was afraid I would break something :P

DNA:
Two outputs based on environment was hard to test, so I used the same URL generation used in prod in dev/tests, so we can see the same output in any environment. 

I also added 2 tests to check if the file extension are transformed correctly, so we can prevent further errors. Seems to be working fine now

DNA:
Fixed the manga path that I let slide on the previous commit. The code got bigger than I initially thought, but now it's should handle image sizes and paths better. It can be better optimizated, but that's is another task :smile:

app/models/manga.rb
```diff
@@ -86,16 +86,31 @@ def canonical_title
     romaji_title
   end
 
-  def poster_image_thumb
+  def poster_image_thumb(size = :large)
```
NuckChorris: If this method is identical between `Anime` and `Manga`, it should probably be broken into a concern (especially with these changes adding so many LOC)

DNA: I thought so… I wasn't sure to refactor so much since it was my first time working in the code but I'll make a proper refactor when I finish the api search :)

DNA:
Ok, rebased this PR with master. Now that `poster_image_thumb` are on `MediaConcern`, it should solve our path problems :)

DNA:
Almost 2 months after my first contribution to hummingbird, my first PR are ready to go! :dancer:

vevix:
Did we do any benchmarking on the monkeypatch?

NuckChorris:
LGTM!

Issue #310, opened 2014-11-21
[API] Genres not included in anime objects in lists


xiprox:
Hello,

I'm working on some kind of an application that utilizes the hummingbird API (v1). When I load anime details, say, from library, I directly pass the existing anime object to the details page instead of passing the id and fetching the details from the server once again. This is great for performance, however, I found out that genres aren't included in an anime object found in feed, timeline, or library. I don't want to download a whole anime object just for genres.

I checked the source code for the API and I see that there is a boolean check for including genres. This boolean is passed as false whenever an anime object is generated as a part of a list. Is there a reason for this? Is it possible to include genres in anime objects returned in library or timeline calls?

Thanks.

PR #533, opened 2015-05-05
Add genres to anime listings on API V1

DNA:
This PR add genre data to animes on API V1, taking care to not incur in N+1 problems, as solicited on #310. Here is the affected endpoints:

- /timeline
- /users/:user_id/library
- /users/:user_id/feed
- /search/anime/
- /users/:user_id/favorite_anime

BTW, `/users/:user_id/favorite_anime` was with a N+1 problem when loading animes entries. This was also corrected

### DIFF
--- a/app/api/api_v1.rb
+++ b/app/api/api_v1.rb
@@ -57,7 +57,7 @@ class API_v1 < Grape::API
       end
     end
 
-    def present_watchlist(w, rating_type, title_language_preference)
+    def present_watchlist(w, rating_type, title_language_preference, include_genres=true)
       {
         id: w.id,
         episodes_watched: w.episodes_watched,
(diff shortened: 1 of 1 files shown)


NuckChorris:
Is there anywhere that _doesn't_ include genres?  Does it really need to be optional?  Otherwise :+1: looks great

DNA:
To be honest, I went though the hole code and haven't see an endpoint without genres… I think we can also remove the `include_genres=true` parameter on those methods… o_o"

NuckChorris:
:+1: on ripping out the `include_genres` parameter.  Options are, after all, the devil.

DNA:
Done! One parameter less :smile:

NuckChorris:
* [x] Net minus 6 lines
* [x] Makes APIv1 code better
* [x] Fixes an N+1 in APIv1
* [x] Simplifies the parameters in N+1

Hell yeah

Issue #352, opened 2015-01-05
New stuff needed for APIv2


vikhyat:
* [ ] Get anime by season
* [ ] Network and airing time + day of week. We don't currently have this but TheTVDB exposes it, so we should start saving it

nizlumina:
Great to see this! Currently I'm scraping parts of Hummingbird for this. Not a pretty solution so far. After hitting a bit of a stump with lots of the other anime-based-sites API for this feature, I really have to say (and humbly *stress*) that an official API for this is very much needed.

NexPB:
Franchise anime ids would be helpfull as well.

tunderdomb:
+1 I'd love to use it.
For reference: search by season, year of release, airing status (finished/airing/upcoming) would be a great addition to the API.

DNA:
In this anime by season search, how would be the endpoint? I was thinking of something like `/anime/?season=fall&year=2015`

NuckChorris:
I believe it would be `/api/v2/anime?query=params&in=here` where the query parameters are representative of the different fields.  The [JSONAPI](http://jsonapi.org/) spec may have something to say on the matter.

DNA:
Yes, JSONAPI have a recomendation on this matter: [Recommendations for Filtering](http://jsonapi.org/recommendations/#filtering). Based on this, It would be like this:

    GET /anime?filter[season]=fall&filter[year]=2015

Can I work on this API search? Or is it task already taken? :)

NuckChorris:
Feel free to take it.  If you have any questions you can hit up our Slack (email josh@hummingbird.me to get an invite) and we're all happy to help

PR #537, opened 2015-05-09
API V2 anime search

DNA:
Well, this is just the first commit in a WIP to add a search functionality to the API V2, related to issue #352 

Probably the API users will like to search a lot of fields, so I'm incapsulating it in an specific model, that handle validation and apply search parameters into the model. In this first commit we can search for a name with the following query:

`/anime?filter[name]=anime_name`

But we can also add some parameters and test validation. It should show the same result as we search the same name on the site (`:full_search` scope). I'll write some tests after I'm sure this is the right path to go. 

As this is a new feature, I'd like to hear your opinions as I go, since this is the first "big thing" I do here on hummingbird :)


NuckChorris:
For a "first big thing" you chose a great task :dancer:

Few things.  First of all, I _love_ that you combined searching and browsing into a single API.  The distinction was always one of my pet peeves, and this is a perfect solution.

Second, this is some beautifully clean code and your use of `ActiveModel::Validators` to validate the query parameters is brilliant and something I would never have thought of.

That said, I have some misgivings about using a `model` to encapsulate this.  It's not really a model, per se.  What about a PORO `service` object which mixes in `ActiveModel::Validators` directly instead of dragging in the entirety of `ActiveModel::Model`?

Besides that, there's a few style things I'll put line comments on (and which I should _probably_ add to `CONTRIBUTING.md`) — we mostly follow-ish bbatsov's style guide

app/controllers/api/v2/anime_controller.rb
```diff
@@ -2,6 +2,21 @@ module Api::V2
   class AnimeController < ApiController
     caches_action :show, expires_in: 1.hour
 
+    def index
+      if params.key? :filter
+        search = AnimeSearch.new params[:filter]
```
NuckChorris: Methods which aren't part of a DSL should generally have parentheses for their call (this is a vague distinction, I know)

app/models/anime_search.rb
```diff
@@ -0,0 +1,40 @@
+class AnimeSearch
+  include ActiveModel::Model
+
+  SHOW_TYPES = %w(OVA ONA Movie TV Special Music)
+
+  SEASONS = %w(summer fall winter spring)
+
+  attr_accessor(
```
NuckChorris: Big lists of symbols like this should be aligned, and since `attr_accessible` is a DSL method, it shouldn't have parentheses.  Like so:

```ruby
attr_accessor :name,
              :slug,
              :show_type
```

app/models/anime_search.rb
```diff
@@ -0,0 +1,40 @@
...
+    :age_rating,
+    :episode_count,
+    :started_airing,
+    :finished_airing,
+    :genres,
+    :season,
+    :year
+  )
+
+  with_options({allow_blank: true}) do |item|
```
NuckChorris: For parameter-hashes, omit the curlies

app/models/anime_search.rb
```diff
@@ -0,0 +1,40 @@
...
+    item.validates :show_type, inclusion: SHOW_TYPES
+    item.validates :season, inclusion: SEASONS
+    item.validates :started_airing , :finished_airing, format: { with: /\d{4}-\d{2}-\d{2}/, message: 'date must be formatted as YYYY-MM-DD' }
```
NuckChorris: Wrap to 80 columns

app/models/anime_search.rb
```diff
@@ -0,0 +1,40 @@
...
+    item.validates :year, format: { with: /\d{4}/, message: 'year must be formatted as YYYY' }
+  end
+
+  def apply anime
+
```
NuckChorris: Shouldn't be space inside the method, but I assume that's intentional.  Also the method parameters should be in parentheses for declarations: `def apply(anime)`

DNA:
First of all, thanks for the nitpicking on the code style. I'm somewhat new to rails, and I'm sure this will help me improve my code :)

And about the other comments:

1 - Yes, I also hate to have separated endpoints to do search. When we have some broad searchz (like you guys have on site searching for anime mangas and users) I think it's ok, but for an API having the option to search in the same endpoint is a must.

2 - I choose to use a model because *in my view*, search ***is a business logic***, even if not table related. Giving the needed data to our user is important, and preventing that the user break our search sending wrong data to our server is algo business logic in some way.

But that's my view based on what I learnt from other sources, if you guys think that's better to put on other place, OK for me. I just thought that it would be better to let it be an model, so we can concern it and improve as we extend it for other models. Having other approaches is the objective from creating this WIP in the first place :)

NuckChorris:
>I choose to use a model because in my view, search is a business logic

I think the concept that "all business logic belongs in the models" is kind of an overgeneralization.  All business logic *should* be encapsulated into objects and removed from the controllers and views, but sometimes a model just isn't the right type of object.  For example, to integrate the Story and Substory and Notification models into a coherent system, we have the NewsFeed service.

We can still do concerns with service objects too, because concerns are just fancy mixins, though ideally we'd have a generalized interface between Anime and Manga which makes the unshared code very minimal.

After consulting with #RubyOnRails IRC, I think the `service` object is probably the right approach here, and probably just involves moving your file from `app/models` to `app/services`, replacing `ActiveModel::Model` with `ActiveModel::Validators`, and (if Rails doesn't autoload service objects) adding a `require 'services/anime_search'` to the controller file

DNA:
Ok, no problems. I'll check the other services and move the search there on the next commit, following the same parameters/standards :smiley:

NuckChorris:
:thumbsup: There's not much of a "standard" for the service objects, they're really just domain objects which work in the ways suited best to their domain.

app/services/anime_search.rb
```diff
@@ -0,0 +1,59 @@
...
+  def apply_filters anime
+    anime = anime.full_search(name) unless name.nil?
+    anime = anime.where(slug: id) unless id.nil?
+    anime = anime.where(show_type: show_type) unless show_type.nil?
+    anime = anime.where(genres: { name: genres.split(',') }).joins(:genres) unless genres.nil?
+    anime = anime.where(age_rating: age_rating) unless age_rating.nil?
+    anime = anime.where(episode_count: episode_count) unless episode_count.nil?
+    anime = anime.where('started_airing_date > ?', started_airing) unless started_airing.nil?
+    anime = anime.where('finished_airing_date < ?', finished_airing) unless finished_airing.nil?
+    anime = anime.where('EXTRACT(MONTH FROM started_airing_date) in (?)', SEASONS_MONTHS[season.to_sym]) unless season.nil?
```
NuckChorris: Might be good to change these more complex ones into named scopes on the model — `Anime.season(:june, :july)` or something

app/controllers/api/v2/anime_controller.rb
```diff
@@ -2,6 +2,21 @@ module Api::V2
...
+      if params.key? :filter
+        search = AnimeSearch.new(params[:filter])
+        if search.valid?
+          anime = search.apply_filters(anime)
+        else
+          return render json: search.errors.full_messages
+        end
+      end
+
+      render json: anime.page(params[:page])
```
NuckChorris: Kaminari requires that an `order` be used on all paginated things.  You should also probably specify a `per`.

app/services/anime_search.rb
```diff
@@ -0,0 +1,59 @@
...
+                     message: 'date must be formatted as YYYY-MM-DD'
+                   }
+    item.validates :year,
+                   format: {
+                     with: /\d{4}/,
+                     message: 'year must be formatted as YYYY'
+                   }
+  end
+
+  def apply_filters anime
```
NuckChorris: `def apply_filters(anime)`

app/services/anime_search.rb
```diff
@@ -0,0 +1,59 @@
...
+                :finished_airing,
+                :genres,
+                :season,
+                :year
+
+  with_options(allow_blank: true) do |item|
+    item.validates :show_type, inclusion: SHOW_TYPES
+    item.validates :season, inclusion: SEASONS
+    item.validates :age_rating, inclusion: AGE_RATINGS
+    item.validates :started_airing ,
```
NuckChorris: Got a space before your comma here

app/services/anime_search.rb
```diff
@@ -0,0 +1,59 @@
...
+                     with: /\d{4}/,
+                     message: 'year must be formatted as YYYY'
+                   }
+  end
+
+  def apply_filters anime
+    anime = anime.full_search(name) unless name.nil?
+    anime = anime.where(slug: id) unless id.nil?
+    anime = anime.where(show_type: show_type) unless show_type.nil?
+    anime = anime.where(genres: { name: genres.split(',') }).joins(:genres) unless genres.nil?
```
NuckChorris: Won't this cause an N+1 when the filter doesn't include genres?

app/controllers/api/v2/anime_controller.rb
```diff
@@ -2,6 +2,21 @@ module Api::V2
   class AnimeController < ApiController
     caches_action :show, expires_in: 1.hour
 
+    def index
+      anime = Anime.preload(:genres)
+
+      if params.key? :filter
```
NuckChorris: Since this isn't a DSL method, it should probably be `params.key?(:filter)`

NuckChorris:
Looking great :thumbsup: Probably should do tests for it, so we get off on the right (tested) foot for APIv2

app/services/anime_search.rb
DNA: Agreed. And if possible, put a "preload(:genre)" on a default scope to stop worrying about N+1 :smile:

NuckChorris: :+1: agreed.  Feel free to send a PR moving things from preloads in the controller to `default_scope`s

app/controllers/api/v2/anime_controller.rb
DNA: Worked fine without `order`, so I didn't put one. And can't we put this `per()`method somewhere were we don't need to call it everytime?

app/services/anime_search.rb
DNA: No, because I pass `Anime.preload(:genres)` to `apply_filters`. Also didn't put it inside the method so it will still preload with no filters

app/controllers/api/v2/anime_controller.rb
NuckChorris: It doesn't throw any errors without `order`, but the behavior is undefined — database offsets don't make sense without an order and can potentially return random results.

As for the `per`, we should be able to use a `default_scope`

app/controllers/api/v2/anime_controller.rb
DNA: You got me here, I don't know. I don't usually see people using parenthesis with `?`methods… o_o"

NuckChorris: [bbatsov is vague](https://github.com/bbatsov/ruby-style-guide#no-dsl-parens), but seems to lean towards including parens.  GitHub's style guide says nothing on the matter.

From a quick run through `rubocop`, I'm gonna side with GitHub and say "fuck it who cares" for predicate methods

DNA:
Ok, that was a lot of changes here! Some "complex" searches were moved into Anime model. I also moved some code around grouping by it's usage (first includes, then constantes, relations, scopes, etc…) and finally, I've run rubocop to remove some problems in the files. If some code style seems off, it could be it. 

Now I'm moving to testing the api, suggestions and tips are really welcome :smile:

app/models/anime.rb
```diff
@@ -260,105 +335,103 @@ def self.create_or_update_from_hash(hash)
...
   end
 
   def status
-    # If the started_airing_date is in the future or not specified, the show hasn't
-    # aired yet.
-    if started_airing_date.nil? or started_airing_date > Time.now.to_date
-      return "Not Yet Aired"
+    # If the started_airing_date is in the future or not specified,
+    # the show hasn't aired yet.
+    if started_airing_date.nil? || started_airing_date > Time.zone.now.to_date
```
NuckChorris: I don't think we actually need to use `Time.zone` since we work *only* in UTC.  The only part which deals with timezones is the Ember app

app/models/anime.rb
```diff
@@ -260,105 +335,103 @@ def self.create_or_update_from_hash(hash)
...
       end
     end
   end
 
   before_save do
     # If episode_count has increased, create new episodes.
-    if self.episode_count and self.episodes.length < self.episode_count and (self.episodes.length == 0 or self.thetvdb_series_id.nil?)
-      (self.episodes.length+1).upto(self.episode_count) do |n|
-        Episode.create(anime_id: self.id, number: n)
+    if episode_count &&
```
NuckChorris: This seems wrapped way too early, is this 80 columns?

app/models/anime.rb
```diff
@@ -41,81 +41,136 @@
...
+
+  scope :by_year, ->(year) {
+    where('EXTRACT(YEAR FROM started_airing_date) = ?', year)
+  }
+
+  scope :by_started_date, ->(date) { where('started_airing_date > ?', date) }
+
+  scope :by_finished_date, ->(date) { where('finished_airing_date < ?', date) }
+
+  scope :order_by_rating, lambda {
```
NuckChorris: Might be best to put this in the `BayesianAverageable` concern

app/models/anime.rb
```diff
@@ -126,82 +181,89 @@ def self.sfw_filter(current_user)
...
+  def alternate_title(preference = '')
+    if preference.class == User
+      preference = preference.title_language_preference
     end
 
-    if title_language_preference and title_language_preference == "romanized"
-      return alt_title
-    elsif title_language_preference and title_language_preference == "english"
-      return (alt_title and alt_title.length > 0) ? title : nil
+    if preference == 'english' || english_canonical
```
NuckChorris: I _think_ you can access boolean fields as if they were predicate methods so `english_canonical?` might be more conventional

app/models/anime.rb
DNA: Ok! Another thing to change on rubocop?

app/models/anime.rb
DNA: the hole line is bigger than 80. If you split it, you need to align every condition (at least that was the only way to make it work)

app/models/anime.rb
DNA: Agreed. Do I rebase and make this change, or do it after merging this PR?

NuckChorris:
The Anime refactoring looks amazing! I'll have to pull this down and poke around to make sure nothing is obviously broken, but this is a great improvement :dancer:

app/models/anime.rb
DNA: I tested on irb and rails console, but didn't work. I think it need to be explicity defined in the model

app/models/anime.rb
NuckChorris: Hmm.  So we can't just wrap it at 80 at the second `&&`?  :/ I'll have to check if there's a flag in rubocop for that

DNA:
Yes, this PR are getting bigger than expected. I'll just focus on the tests so we can merge this before It gets to far from the code on master.

app/models/anime.rb
NuckChorris: Feel free to rebase and make the change or just make a commit in this PR

NuckChorris:
Good plan.  We have enough PRs that have been slow-roasting for months, we don't need another :stuck_out_tongue_winking_eye:

app/models/anime.rb
NuckChorris: Strange, I swear I just used that in my Apps stufff...

(5 minutes of Googling various terms later) Yeah thought so, http://makandracards.com/makandra/607-automagical-methods-for-boolean-attributes-in-activerecord it just needs to be a boolean column

NuckChorris:
Are you still working on the tests or should we move this WIP->ready?

DNA:
Yes, I'm done testing! Just waiting your guys final OK :)
Em 05/06/2015 23:28, "Peter Lejeck" <notifications@github.com> escreveu:

NuckChorris:
It's not perfect, but it's easily better than 98% of our code.  I'd say it's ready.

NuckChorris:
Related, for anyone who wants it: probably-pretty-easy cleanup over in the [AnimeController](https://github.com/hummingbird-me/hummingbird/blob/746450e660/app/controllers/anime_controller.rb#L29-L255) using this.  In fact, the AnimeController is the [worst non-vendor code](https://codeclimate.com/github/hummingbird-me/hummingbird/AnimeController) in the HB codebase.

NuckChorris:
:ballot_box_with_check: Deployed!

Somebody wanna do the honors of adding it to [our APIv2 docs](https://github.com/hummingbird-me/hummingbird/wiki/API-v2-Methods)?

app/controllers/api/v2/anime_controller.rb
```diff
@@ -2,6 +2,21 @@ module Api::V2
...
+        search = AnimeSearch.new(params[:filter])
+        if search.valid?
+          anime = search.apply_filters(anime)
+        else
+          return render json: search.errors.full_messages
+        end
+      end
+
+      render json: anime.page(params[:page]).per(20)
```
NuckChorris: Just found a :beetle: on this that needs fixing before anyone starts using the endpoint: this needs to be `AnimeSerializer.new(anime).as_json` (I don't believe `Api::V2` is required) like below, and we will want to include `meta` for pagination

DNA: I'll fix it by Monday (I'll also check error messages, http standards and
other json api items).

I can also update the docs, I'll just need some proofread for my poor
English :P
Em 07/06/2015 01:14, "Peter Lejeck" <notifications@github.com> escreveu:

NuckChorris: :+1: Sounds fine, I wouldn't be able to get to it any sooner anyways.

PR #538, opened 2015-05-09
Create .rubocop.yml

NuckChorris:
Based on a modified bbatsov to bring it somewhat in line with current practices.

We should make sure future code is up to these standards, and when we feel we won't get spammed, reenable Hound.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,259 @@
+AllCops:
+  RunRailsCops: true
+  DisplayCopNames: false
+  DisplayStyleGuide: true
+  StyleGuideCopsOnly: false
+Style/BlockDelimiters:
+  EnforcedStyle: semantic
+Style/CollectionMethods:
+  PreferredMethods:
+    collect: map
+    collect!: map!
+    inject: reduce
+    detect: find
+    find_all: select
+Style/CommentAnnotation:
+  Enabled: true
+  Keywords:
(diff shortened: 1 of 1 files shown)


DNA:
user$machine> git checkout -b add-rubocop-config --track upstream/add-rubocop-config
    Switched to branch 'track upstream/add-rubocop-config'
    user$machine> rubocop
    358 files inspected, 4630 offenses detected

4630 offenses. Is it normal to expect, or is something that we need improvement? :P

(or I misconfigurated something?)

NuckChorris:
No, it's entirely likely that there's really that many offenses.  Some of the old code is really bad.  There's a reason we haven't enabled Hound.  Well, not since _that one time_ when it spammed our watchers with roughly 100 messages on one PR.

As we write new code going forward, we should make sure it passes rubocop

vevix:
LGTM

(Mobile so no lgtm gif :sob:)

Eschon:
Can I help you with resolving the offenses? If yes where would be a good point to start?

NuckChorris:
@Eschon just run `rubocop` on a file, fix things it reports, and send a PR :D

Don't bother with huge refactors or reducing cyclomatic complexity If rubocop reports that, just low-hanging fruit is fine. Other things can get fixed as we go forward!

If you have any questions, join the slack chat and we can help :)

PR #541, opened 2015-05-10
resolve rubocop offenses for home_controler.rb

Eschon:
I decided to start small by resolving the offenses in home_controler.rb

There is only one left that I didn't touch:
* Do not itroduce global variables.
at `break_counter = $redis.with { |conn| conn.get('break_counter') } || 0`

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -19,7 +19,7 @@ class HomeController < ApplicationController
       user_id: current_user.id,
       recent: true,
       include_private: true,
-      include_adult: !current_user.sfw_filter?,
+      include_adult: !current_user.sfw_filter?
     )
     generic_preload! 'recent_library_entries',
                      ed_serialize(recent_library_entries,
@@ -41,7 +41,7 @@ class HomeController < ApplicationController
   def unsubscribe
     code = params[:code]
     User.all.select { |x| x.encrypted_email == code }
-      .each do |x| x.update_attributes subscribed_to_newsletter: false end
+      .each { |x| x.update_attributes subscribed_to_newsletter: false }
     flash[:message] = 'You have been unsubscribed from the newsletter.'
     redirect_to '/anime'
   end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -3,7 +3,7 @@ require_dependency 'library_entry_query'
 class HomeController < ApplicationController
   def index
     if user_signed_in?
-      redirect_to "/dashboard"
+      redirect_to '/dashboard'
     else
       render_ember
     end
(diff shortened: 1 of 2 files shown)


app/controllers/home_controller.rb
```diff
@@ -38,9 +40,10 @@ def static
 
   def unsubscribe
     code = params[:code]
-    User.all.select {|x| x.encrypted_email == code }.each {|x| x.update_attributes subscribed_to_newsletter: false }
-    flash[:message] = "You have been unsubscribed from the newsletter."
-    redirect_to "/anime"
+    User.all.select { |x| x.encrypted_email == code }
+      .each do |x| x.update_attributes subscribed_to_newsletter: false end
```
NuckChorris: I should change this in rubocop. Since Hummingbird is a big team, having a vague distinction like "do..end for imperative, {} for functional" is probably a bad idea.  I'll switch this back to the standard bbatsov "do..end for multi line, {} for single line"

app/controllers/home_controller.rb
```diff
@@ -19,15 +19,17 @@ def dashboard
       user_id: current_user.id,
       recent: true,
       include_private: true,
-      include_adult: !current_user.sfw_filter?
+      include_adult: !current_user.sfw_filter?,
```
NuckChorris: Is there a rule demanding that there be a trailing comma?  I should probably disable that

NuckChorris:
There's clearly a few things that need adjustment in rubocop's config, so here's a checklist:

 * [ ] block style should be changed from "functional-vs-imperative" to "multiline-vs-singleline"
 * [ ] trailing comma in hash should be disabled (is it enabled?)
 * [ ] Ignore global variables

app/controllers/home_controller.rb
Eschon: Yeah, I thought it was strange but the reason I found for that is to make it easier to add more lines and I thought you left it in with that in mind.

NuckChorris: haha it does provide that but it also makes it unclear where the end of the list is and is a _strange_ style in that nobody uses it.  We should probably disable it.

Eschon:
It shouldn't be a problem for me to do this but I probably can't work on this till Wednesday.

app/controllers/home_controller.rb
Eschon: Ok I can change that back.

NuckChorris:
That's alright, it's not urgent :thumbsup:

app/controllers/home_controller.rb
DNA: It's a common practice to put a comma on the last hash item because if you have to add another value later on, you'll have two lines on the git diff, the line added and the comma on the previous line

NuckChorris: I know it makes sense but I've honestly not seen it in other projects

DNA: In PHP we use it a lot, since it's ARRAYS EVERYWHERE :laughing:

NuckChorris: If we were learning from PHP's style we'd never get anything done! I say we just use the bbatsov style guide default and call it good, so we don't have to [bikeshed](http://bikeshed.org)

Eschon:
Made the suggested changes everything should be alright now.

app/controllers/home_controller.rb
```diff
@@ -38,9 +40,10 @@ def static
 
   def unsubscribe
     code = params[:code]
-    User.all.select {|x| x.encrypted_email == code }.each {|x| x.update_attributes subscribed_to_newsletter: false }
-    flash[:message] = "You have been unsubscribed from the newsletter."
-    redirect_to "/anime"
+    User.all.select { |x| x.encrypted_email == code }
+      .each { |x| x.update_attributes subscribed_to_newsletter: false }
```
NuckChorris: Please tell me I'm reading this wrong and it doesn't _actually_ load all users into RAM just to unsubscribe a single one of them?

M-maybe to avoid timing attacks... against our unsubscribe endpoint?  I just... wtf

NuckChorris:
:+1: LGTM

app/controllers/home_controller.rb
Eschon: Yeah looks like that's the case. Maybe this should be changed but I think I don't know enough about your code and Rails yet to do that.

PR #543, opened 2015-05-10
Polished the "About ..." formular and fixed #532

### LINKED ISSUE #532: Favourite button misaligned on manga pages.
screen shot 2015-05-06 at 1 44 46 am

Most likely because we use a different poster image size on manga (why?)

--- 

Side note:

We should show character images by default on manga pages (as it assumes their is a staff member image).

simonknittel:
Before:
before

After:
after

### DIFF
--- a/app/assets/stylesheets/partials/anime-page.css.scss
+++ b/app/assets/stylesheets/partials/anime-page.css.scss
@@ -84,7 +84,7 @@
   @media (max-width: 860px) {
     width: 100%;
   }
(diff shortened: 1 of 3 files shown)


JoshFabian:
This looks much nicer!

wopian:
Made me notice that the inputs aren't in-line with the textarea.

NuckChorris:
Probably cause of bootstrap column gutters

PR #546, opened 2015-05-12
Add faker and factories

NuckChorris:
Extracted from #461, since that PR isn't ready but this is useful for other purposes in testing (namely that I want to use it for my `App` tests in the OAuth2 stuff)


vevix:

PR #549, opened 2015-05-14
Run hummingbird in Docker

MiLk:
This PR add a `Dockerfile` and a `docker-compose.yml` to run hummingbird in Docker with docker-compose.

## Requirements

* [Docker](https://docs.docker.com/installation/)
* [docker-compose](https://docs.docker.com/compose/install/)

### Build the container

```bash
docker-compose build
```

### Start the application

```bash
docker-compose up web
```

#### Native docker

```bash
xdg-open http://localhost:$(docker-compose port web 3000 | cut -d':' -f2)
```

#### Boot2docker

```bash
open http://$(boot2docker ip):$(docker-compose port web 3000 | cut -d':' -f2)
```

_____________________________________

Todo:
- [x] Move the entrypoint script from `/script` to `/bin`
- [x] Use relative paths inside the entrypoint script
- [x] Install node 0.12
  - [x] Install nvm
- [x] Execute pgsql tasks only on first run
  - [x] Load sql dump from https://github.com/hummingbird-me/hummingbird/pull/556/files#diff-42a8065bc1d626f4d9816ebce05e3f71R35

_____________________________________
Initial discussion:

I'm trying to run Hummingbird in Docker, because I don't want to install redis, postgres and ruby on my machine.

I had a hard time with some dependencies, but the app now seems to start correctly.

However, I got the following error when accessing the app:
2015-05-14-215358_1056x811_scrot

The docker container is built following [these instructions](https://github.com/hummingbird-me/hummingbird/wiki/Setting-up-a-development-environment#ubuntudebian-linux) (Thanks for the update done yesterday).

What should I do?


config/database.yml
```diff
@@ -16,3 +11,11 @@ production:
   host: <%= ENV['POSTGRES_HOST'] %>
   port: 5432
   pool: 50
+
+development:
+  adapter: postgresql
+  host: <%= ENV['POSTGRES_PORT_5432_TCP_ADDR'] %>
```
NuckChorris: This should probably revert back to the previous defaults when not in docker, for the sake of not destroying in-use development environments or causing trouble for people not using docker.  Might be achievable via defaults in `.env`?

NuckChorris:
Check that the `frontend/node_modules` directory is populated.  We have this saved as a dependency, so I'm not sure why it's telling you to save it as a dependency.

config/database.yml
MiLk: Yes, still WIP.

MiLk:
I'm going further.

Looks like every time I'm doing a request, the app tries to rebuild the ember app, and the worker timeout.
```
web_1      | 02:57:42 unicorn.1 | Future versions of Ember CLI will not support v0.10.25. Please update to Node 0.12 or io.js.
web_1      | 02:57:42 unicorn.1 | version: 0.2.3
web_1      | 02:57:51 unicorn.1 | 0.2.4
web_1      | 02:57:51 unicorn.1 |
web_1      | 02:57:54 unicorn.1 | Could not find watchman, falling back to NodeWatcher for file system events.
web_1      | 02:57:54 unicorn.1 | Visit http://www.ember-cli.com/#watchman for more info.
web_1      | 02:57:55 unicorn.1 | BuildingD, [2015-05-15T02:57:55.332403 #142] DEBUG -- : waiting 11.0s after suspend/hibernation

web_1      | 02:59:20 unicorn.1 | Building.Building..Building...BuildingBuilding.Building..Building...BuildingBuilding.Building..============================= WARNING! =============================
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 |   Seems like Ember frontend application takes more than 15
web_1      | 02:59:20 unicorn.1 |   seconds to compile.
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 |   To prevent race conditions consider adjusting build timeout
web_1      | 02:59:20 unicorn.1 |   configuration in your ember initializer:
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 |     EmberCLI.configure do |config|
web_1      | 02:59:20 unicorn.1 |       config.build_timeout = 20 # in seconds
web_1      | 02:59:20 unicorn.1 |     end
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 |   Alternatively, you can set build timeout per application like this:
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 |     EmberCLI.configure do |config|
web_1      | 02:59:20 unicorn.1 |       config.app :frontend, build_timeout: 20
web_1      | 02:59:20 unicorn.1 |     end
web_1      | 02:59:20 unicorn.1 |
web_1      | 02:59:20 unicorn.1 | ============================= WARNING! =============================
web_1      | 02:59:23 unicorn.1 | Building...E, [2015-05-15T02:59:23.395319 #142] ERROR -- : worker=0 PID:225 timeout (21s > 20s), killing
web_1      | 02:59:23 unicorn.1 | E, [2015-05-15T02:59:23.409353 #142] ERROR -- : reaped #<Process::Status: pid 225 SIGKILL (signal 9)> worker=0
web_1      | 02:59:24 unicorn.1 | I, [2015-05-15T02:59:24.518769 #243]  INFO -- : worker=0 ready
```

I tried to build the ember app before running the app, but it doesn't seem to change anything.
I tried to change the `build_timeout` setting, but it doesn't seem to change anything.
What should I do?

NuckChorris:
This is gonna sound bad, and it kind of is, but the only answer I have found is "mash reload and pray"



—Peter

MiLk:
It takes 2m30s to build the ember app on my laptop at work :/ (5400rpm HDD, OSX)
Will try again at home tonight (SSD, native docker, ...).

NuckChorris:
Can confirm, I get similarly bad build times.  Broccoli made the horribly misinformed decision to use the disk for temp space and now they're paying a hefty price for it.

Luckily, they use incremental rebuilds so later builds aren't so bad, once the pump is primed.  But getting the pump primed is an awful price.  One of the things we really want to do is split up the Ember and Rails apps so that we can ditch `ember-cli-rails` and avoid these problems caused by the combination of a server timeout and a long build.

Sadly, we're a good ways away from being able to do that.  We need to remove basically all the legacy Rails pages and change the way we do preloading and authentication.

We also need an encapsulated development environment so we can freely add dependencies!  So this PR is a big step in the right direction, especially if we couple it with a wrapping Vagrantfile.

MiLk:
It's getting better.
~8s to build on my computer at home.

NuckChorris:
Just want to add here that I've released an updated [dump.sql](https://slack-files.com/T025CNWM0-F04TCL9QX-42e5a92d04) for development with.

MiLk:
I've improved the way the postgresql settings are set.
But I don't know how to manage redis config.
Is there a way to define a variable in the `.env` file which will be used only if there is nothing in the environment?

NuckChorris:
You can probably put defaults for development sans docker in `.env` and use `ENV` in Docker or whatever to override it, right?

MiLk:
I think it's now working.
I have updated the first message to explain how to use it.

NuckChorris:
You have no idea how friggin' hyped I am for this.  Anyways, all I have left are some questions about `scripts/entrypoint.sh`:

1. Why are some of the env things in `.dockerenv` but others in `entrypoint.sh`?
2. Should it be in `bin` instead of `scripts`?  I think `script` became `bin` somewhere around Rails 3 or 4...
3. Is it possible to make the script use relative directories so we can run it in a different environment?
4. Should it automatically pull down the dump.sql like Epictek's bootstrap.sh? See https://github.com/hummingbird-me/hummingbird/pull/556/files#diff-42a8065bc1d626f4d9816ebce05e3f71R35 (but note that you can actually just pipe curl into psql instead of using a temp file)
5. Is the `node` at least a `0.12`?  Otherwise ember-cli gets bitchy
6. How hard will it be, down the road, to add ElasticSearch to this?  Because I want to install that for a better search eventually

MiLk:
1. I put in `.dockerenv` only the variables which *MUST* override the ones in `.env`. I should maybe move everything to the `.dockerenv` file.
2. Tbh, I really hate ruby (one of the reason why I want to run that in docker and not on my computer), so I have no idea what are the conventions for Rails, I will move that to the `bin` directory.
3. What do you mean? Running `./bin/entrypoint.sh bundle exec foreman start` to start your app?
4. I can do that.
5. No, I can install it with nvm instead of the ubuntu packages.
6. It will be really easy. Just add https://registry.hub.docker.com/_/elasticsearch/ to the `docker-compose.yml` file, and do the config magic as for postgresql.
7. Usually it's `/docker-entrypoint.sh`, it can become `/bin/docker-entrypoint.sh`.

NuckChorris:
1. That would probably be 90% of addressing 3
2. hah, that works.  Yeah bin is probably the right place
3. What I'm referring to are the use of `/opt/hummingbird/` explicitly when perhaps they could be relative paths — between this and 1, it would probably be possible to use this on any system (even without Docker) to bootstrap a dev server
4. :thumbsup:
5. Probably wise.  ember-cli is giving a deprecation warning for pre-0.12 versions, something about "future versions will not support Node.js prior to 0.12"
6. Fucking yes I love it!
7.If that's the convention it's great, I'm just wondering if this isn't useful for a lot more than just Docker; it seems to me like this same script would work just as well on OSX or in a normal Ubuntu install for handling half the install automatically

MiLk:
For 3 + 7. we can maybe use a Makefile.

A simple `make start` will call the entrypoint with the right parameters for non-docker environments.
And something like `make db` could take care of all the rake db commands.

Anyway, I'm a bit busy today, but I should be able to make the discussed changes tomorrow night.

NuckChorris:
Well, we'd just use `rake` for that since we already have `rake`.  But I'm not sure that's even necessary.  Won't the script be run as `/opt/hummingbird/script/entrypoint.sh`, so wouldn't it be possible to just read off the script's basename, make it absolute, resolve up a directory, and use that as the root?

MiLk:
Yes, but look at the last line of the script, it executes anything you pass as a parameter.
It doesn't execute the app itself.

Using this script as it, outside of docker, will require a very long command: `./script/entrypoint.sh bundle exec foreman start`.

NuckChorris:
I don't think the length of the command would be a problem, `bundle exec foreman start` is already pretty long, and this would be such an improvement in the non-docker environment.

That said, I guess it doesn't really matter: I'd like to get everybody moved to Docker in dev ASAP, so we can easily add dependencies and split components as we see fit.  It might even be beneficial that it isn't general, in fact.  Motivation for devs to use the Docker option.

MiLk:
I can't move the pgsql variables away from the entrypoint script, because they are used by the rake tasks.
The `.dockerenv` file is used only when the app is started by foreman.

I've done some changes.
I still need to work on node and the pgsql related tasks.

DNA:
In the matter of node 0.10 vs 0.12, a friend of mine installed io.js instead of node when he needed 0.12 version at his work and worked just fine. If the linux repositories are ready for this, I think it's a good workaround, and we can switch back to node on 0.13, when both projects will merge again

DNA:
Also, dunno if you guys already saw that, but just for reference: https://github.com/nodesource/distributions

MiLk:
The easiest way to install node is using nvm.
Btw there won't be a node 0.13.
They will take the next major version after the last one from iojs.
It's part of the reconciliation plan.

DNA:
Didn't know about the major version thing… Good to know, it's was about time :)

MiLk:
I should have addressed all the discussed issues.
よろしくお願いします。

NuckChorris:
So if I'm not mistaken, we have to rebuild this Docker container after every change? :/ is there any workaround for that?

MiLk:
https://github.com/MiLk/hummingbird/blob/docker/docker-compose.yml#L11
You have to rebuild it only to update the dependencies if you change them (gem, npm modules, bower, ...).

NuckChorris:
Ah yes, I see it now in `docker-compose.yml`! Well that's the only concern I had left

MiLk:
Thanks :)

NuckChorris:
[Added to docs](https://github.com/hummingbird-me/hummingbird/wiki/Setting-up-a-development-environment#docker-recommended).  I'll probably get switched over to this soon (since I have ongoing branches it'll take me a little while lol)

PR #550, opened 2015-05-15
Apps page updates

NuckChorris:
This is the precursor to the OAuth2 stuff.

This adds fields to the `App` model which we'll need for OAuth2 (redirect uri, logo, etc.) and updates the UI to handle the increase in fields — namely by switching to a separate edit/create route instead of using an inline create card.  Oh, and you can edit apps now.  Cool stuff.

Along the way I added a `save_and_render` helper which will save a model and handle responding with either the model or the errors in ED form and made it so `ember_action` can now be called without a block to just hand off to Ember without preloading.

On the testing side of things I added `guard-minitest` so we can now just start `guard` and it'll monitor and run the tests, tweaked the defaults on `json_expressions` so it's "loose" by default (ie, we don't have to spec the entire object, just what matters to our test), and added `assert_json_body` along the lines of `assert_preloaded` and our other assertion shorthands.

### DIFF
--- a/db/migrate/20150427224114_add_o_auth_to_app.rb
+++ b/db/migrate/20150427224114_add_o_auth_to_app.rb
@@ -1,0 +1,9 @@
+class AddOAuthToApp < ActiveRecord::Migration
+  def change
+    add_column :apps, :redirect_uri, :string
+    add_column :apps, :homepage, :string
+    add_column :apps, :description, :string
+    add_column :apps, :privileged, :boolean, null: false, default: false
+    add_attachment :apps, :logo
+  end
+end
--- a/frontend/app/templates/apps/edit.hbs
+++ b/frontend/app/templates/apps/edit.hbs
@@ -13,6 +13,7 @@
       </div>
       <div class="panel-body">
         <form class="form-horizontal">
+          {{!-- TODO: add logo upload button and display logo
           <div class="form-group">
             <label for="app-logo" class="col-sm-4 control-label required-label">Logo</label>
             <div class="col-sm-8">
@@ -20,6 +21,7 @@
               {{view "file-upload" action="logoSelected" id="app-logo"}}
             </div>
           </div>
+          --}}
 
           <div class="form-group {{if errors.model.name 'has-error'}}">
             <label for="app-name" class="col-sm-4 control-label required-label">Name</label>
@@ -50,6 +52,7 @@
             </div>
           </div>
 
+          {{!-- TODO: add database stuff for this
           <div class="form-group">
             <label class="col-sm-4 control-label required-label">Access</label>
             <div class="col-sm-8">
@@ -68,7 +71,9 @@
               </span>
             </div>
           </div>
+          --}}
 
+          {{!-- NOTE: enable this when we add usage tracking + dynamic /apps
           <div class="form-group">
             <label class="col-sm-4 control-label required-label">Visibility</label>
             <div class="col-sm-8">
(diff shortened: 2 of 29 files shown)


db/migrate/20150427224114_add_o_auth_to_app.rb
```diff
@@ -0,0 +1,9 @@
+class AddOAuthToApp < ActiveRecord::Migration
+  def change
+    add_column :apps, :redirect_uri, :string
+    add_column :apps, :homepage, :string
+    add_column :apps, :description, :string
+    add_column :apps, :privileged, :boolean, null: false, default: false
```
NuckChorris: I should note that the `privileged` field is there to specify that an app has been vetted by us and deemed "trustworthy enough" to give access to spooky-scary APIs like the "resource owner password" flow or the "global" scope.

frontend/app/templates/components/app-card.hbs
```diff
@@ -0,0 +1,45 @@
+<div class="app-card">
+  <div class="app-inner">
+    <h3 class="app-title">
+      {{app.name}}
+      {{#link-to 'apps.edit' app class='btn btn-default app-type pull-right'}}
```
NuckChorris: I just realized I used single quotes in HBS here but double quotes in `apps/edit.hbs` — which is the preference?  I can quickly swap 'em all

frontend/app/templates/apps/edit.hbs
```diff
@@ -0,0 +1,129 @@
...
+          <div class="form-group {{if errors.model.description 'has-error'}}">
+            <label for="app-desc" class="col-sm-4 control-label required-label">Description</label>
+            <div class="col-sm-8">
+              {{input type="text" class="form-control col-sm-8" id="app-desc"
+                      value=model.description placeholder="Description"}}
+            </div>
+          </div>
+          {{show-errors if=model.description fieldName="Description"
+                        clientErrors=errors.model.description
+                        serverErrors=model.errors.description}}
```
NuckChorris: I know what you're thinking: "nuck wtf why do we have separate clientErrors and serverErrors" — the answer is dockyard/ember-validations#122

Once they fix that and we get on 2.0.0, this whole thing becomes a _lot_ simpler.

frontend/app/templates/components/app-card.hbs
vevix: DBL for HBS

NuckChorris: :thumbsup: roger that, I'll fix these in morning

db/migrate/20150427224114_add_o_auth_to_app.rb
DNA: Wouldn't it be better to change to something like "access_status" and use a string field, so we can use it for more statuses? (something like silver/gold app status, each one with more permissions)

I don't know if I'm optimizing prematurely, but I think it's better to ask this on a PR rather than having to alter the table in the future

DNA:
I think it need just some tests to check the model against invalid input data. And I don't know much about ember, but seems fine :shipit:

db/migrate/20150427224114_add_o_auth_to_app.rb
NuckChorris: I don't think there's a need. It's not like it's something these apps pay for, it's just a bit of increased scrutiny. There's no more ranks to it except vetted and unvetted. If circumstances change, a migration is easy, so there's no need to think that far ahead

NuckChorris:
I don't generally write tests for built-in validations. They can't catch anything and just force you to repeat yourself.

I'm more interested in maybe testing logic in the controllers and components. Perhaps the computed properties in the `show-errors` component?

vevix:
Pulled it down to test, seems the `name` field on the edit page is in its error state when blank (Throws a JS error too).

NuckChorris:
Yeah I know it's in the error state but I don't think it really matters — ideally the form would be invalid while the field would be valid-until-focused but that's hard to do in ember. Throws a JS error though? That I did not know

NuckChorris:
Fixed both of those :+1: we should be good to go

vevix:
:+1:

Issue #472, opened 2015-03-21
embedded youtube video refresh


sachaw:
after liking a YouTube video, whilst playing in your feed, the video refreshes, causing it to restart.

NuckChorris:
#405

sachaw:
soz, ty anyway

PR #551, opened 2015-05-16
Ember 1.12

vevix:
Moves us over to Ember 1.12.

- [x] Instance Initializers
- [x] Get/Set Computed Properties
- [x] Fixed "tests"

---------

###### Instance Initializers (Fastboot :+1:)

> In Ember.js 1.12 application boot is separated into two phases:

> * Application initializers run. At this phase of boot, the goal of initializers should be to register dependencies and injections. These initializers are doing work that is shared across all FastBoot requests, and therefore should not create instances. This phase runs once.

> * Instance initializers run next. This is the right time to do work that is specific to each FastBoot request. You can create instances and modify their state here. This phase runs when the browser application runs, for each integration test, and for each FastBoot request.

###### Computed Properties

Old:
```javascript
{
  foo: function(key, value) {
    // setter
    if (arguments.length > 0) { 
      this.set('bar', value);
    }

    // getter
    return this.get('bar');
  }.property('bar')
}
```

New: 
```javascript
{
  foo: Ember.computed('bar', {
    get() { 
      return this.get('bar');
    },

    set(key, value) {
      this.set('bar', value);
    }
  })
}
```


NuckChorris:
Yuck, the new `Ember.computed` stuff is nasty.  At least the future `@computed('height')` annotation syntax will make it all better!

And even more hype for Ember 1.13 and the Glimmer engine.  Finally the "Like" button will no longer rerender the entire story and reset youtube videos, fixing #472

vevix:
I like that it allows us to start using the shorthand function syntax without a lot of ugly mismatch. I was going to transition our usage to it, but I may hold off until the new `@` syntax.

```javascript
{
   foo() { 
     //...
   }
}
```

NuckChorris:
Shorthand function syntax?  Surely you don't mean fat arrows?

vevix:
Nope, properties shorthand (?), nevermind anyway not really important at the moment.

NuckChorris:
Looks like Ember 1.13 is coming in the next couple weeks, so this might want to also do the changes for that

NuckChorris:
Wooooo Ember 1.13 is out (two weeks ago!)! http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html

DNA:
Isn't better to start moving to Ember 2.0? :P

NuckChorris:
Once we're on Ember 1.13 with no deprecation warnings we'll be able to move to 2.0 without basically any work.  2.0 is pretty much just nuking anything which gives deprecation warnings on 1.13

DNA:
Nice! I hope we can pull this out soon :)

vevix:
:ok_hand: ready to go!

They keep deciding to rewrite how test helpers work, so I have to keep fixing up my test PR. Eventually that can be merged too :sob:

frontend/app/instance-initializers/x-push-state.js
```diff
@@ -1,9 +1,9 @@
 import Ember from 'ember';
 
-var XContentReady = new CustomEvent('XContentReady');
+let XContentReady = new CustomEvent('XContentReady');
```
NuckChorris: We can probably remove XPushState, it's horribly buggy and we have something better now: FastBoot™!

frontend/app/mixins/model-truncated-details.js
```diff
@@ -1,23 +1,25 @@
 import Ember from 'ember';
 
 export default Ember.Mixin.create({
-    truncatedBio: function(key, value) {
-      // setter
-      if (arguments.length > 1) {
+    truncatedBio: Ember.computed('bio', {
```
NuckChorris: Is this new syntax only used for computed setters, or are we just taking it slow transitioning away from `.property()`?

vevix: Yeah I plan to move us to this new Ember-style.

NuckChorris: We can merge this without that, right?  I'd say LGTM, this looks pretty damn solid.  It does pass the "poke test", right?  You poked at it a bit and checked that it wasn't exploded in important places (profiles, feeds, libraries)?

vevix: Yep, I plan to do all the rewrites + cleanup post 1.13.

vevix:
I'm gonna poke at this a bit before I merge, as its been awhile.

vevix:
Poked at it :point_right: :point_left:! Gonna wait for @NuckChorris to sign off on the final commit

NuckChorris:
:+1: LGTM! 🎉🎊🎋💥👑💫🎄🎁

PR #557, opened 2015-05-22
added short redirects for animu and mango

Epictek:


wopian:
http://www.gpforums.co.nz/attachment.php?attachmentid=196925&d=1401840215

NuckChorris:
Correction, wopian:

PR #558, opened 2015-05-22
Fix FireFox Issue w/ Review Layout

wopian:
All browsers but FireFox was happy with it at 39px. FireFox requires an extra pixel. :sob: 

39px:

40px:


JoshFabian:
LGTM!

PR #559, opened 2015-05-22
Hacky workaround for `nil` in user array

NuckChorris:
On josh's profile feed, somehow `nil` got into the user loading array, causing his feed to 500.  This is a hacky workaround for that, which just removes `nil` from the array instead of actually tracking down the root cause.

An edgecase like this isn't worth our time or effort to find the root cause right now.


vevix:

PR #560, opened 2015-05-22
Change "video.tv_episode" to "video.other"

wopian:
`video.tv_episode` is not a valid type according to [The Open Graph protocol documentation](http://ogp.me/#type_video). `video.other` is the only suitable video type as Specials and Music are not tv shows (`video.tv_show`) or episodes (`video.episode`).

---

>video.other
>
>A video that doesn't belong in any other category. The metadata is identical to `video.movie`


NuckChorris:
LGTM

PR #561, opened 2015-05-25
resolve rubocop offenses for users_controller.rb

Eschon:
I resolved a few rubocop offenses again. 
Again there where a few that I didn't know how to fix:
* Assignment Branch Condition size too high (for `def index`, `def follow` and `def update`)
* Method has too many lines (for `def index`, `def follow` and `def update`)
* Perceived complexity to high (for `def index`)

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,7 +1,6 @@
 require_dependency 'user_query'
 
 class UsersController < ApplicationController
-
   def index
     if params[:followed_by] || params[:followers_of]
       if params[:followed_by]
@@ -12,7 +11,7 @@ class UsersController < ApplicationController
       users = users.page(params[:page]).per(20)
       UserQuery.load_is_followed(users, current_user)
 
-      render json: users, meta: {cursor: 1 + (params[:page] || 1).to_i}
+      render json: users, meta: { cursor: 1 + (params[:page] || 1).to_i }
     elsif params[:to_follow]
       render json: User.where(to_follow: true), each_serializer: UserSerializer
     else
@@ -38,10 +37,11 @@ class UsersController < ApplicationController
       return redirect_to user_path(user), status: :moved_permanently
     end
 
-    if user_signed_in? and current_user == user
+    if user_signed_in? && current_user == user
       # Clear notifications if the current user is viewing his/her feed.
-      # TODO This needs to be moved elsewhere.
-      Notification.where(user: user, notification_type: "profile_comment", seen: false).update_all seen: true
+      # TODO: This needs to be moved elsewhere.
+      Notification.where(user: user, notification_type: 'profile_comment',
+                         seen: false).update_all seen: true
     end
 
     respond_with_ember user
@@ -56,23 +56,25 @@ class UsersController < ApplicationController
     if user != current_user
       if user.followers.include? current_user
         user.followers.destroy current_user
-        action_type = "unfollowed"
+        action_type = 'unfollowed'
       else
-        if current_user.following_count < 10000
+        if current_user.following_count < 10_000
           user.followers.push current_user
-          action_type = "followed"
+          action_type = 'followed'
         else
-          flash[:message] = "Wow! You're following 10,000 people?! You should unfollow a few people that no longer interest you before following any others."
+          flash[:message] = "Wow! You're following 10,000 people?! You should \
+                             unfollow a few people that no longer interest you \
+                             before following any others."
           action_type = nil
         end
       end
 
       if action_type
-        Substory.from_action({
+        Substory.from_action(
           user_id: current_user.id,
           action_type: action_type,
           followed_id: user.id
-        })
+        )
       end
     end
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/partials/anime-page.css.scss
```diff
@@ -701,4 +701,3 @@
     line-height: 38px;
   }
 }
-
```
Eschon: I don't know where this change comes from and I couldn't get rid of it.
Somehow this line is there when I few the file on my computer.

app/controllers/users_controller.rb
```diff
@@ -138,17 +139,21 @@ def update
...
     end
 
-    changes = changes.permit(:about, :location, :waifu, :website, :waifu_or_husbando,
-                   :bio, :waifu_char_id, :email, :cover_image, :sfw_filter,
-                   :title_language_preference, :password, :name, :star_rating)
+    changes = changes.permit(:about, :location, :waifu, :website,
+                             :waifu_or_husbando, :bio, :waifu_char_id, :email,
+                             :cover_image, :sfw_filter,
+                             :title_language_preference, :password, :name,
+                             :star_rating)
```
Eschon: I don't really like how I broke the lines here. 
Does someone have a better idea how to do it?

app/assets/stylesheets/partials/anime-page.css.scss
DNA: Some IDEs add an empty line on the end of the file, check if it's the case :)

app/controllers/users_controller.rb
DNA: maybe change the symbols order to use the line space better?

app/assets/stylesheets/partials/anime-page.css.scss
NuckChorris: General policy on Unix is to end all files in `\n` so it could be that your editor decided to remove that.  I recommend installing an EditorConfig plugin for your editor or trying a different editor

app/controllers/users_controller.rb
NuckChorris: Changing order might help, if that doesn't help enough, you might be able to realign somehow or maybe use `*%i[about location waifu website]` (a splat operator over a symbol-array literal) — I've never actually tried that but it _should_ theoretically work

NuckChorris:
> Assignment Branch Condition size too high (for def index, def follow and def update)
> Method has too many lines (for def index, def follow and def update)
> Perceived complexity to high (for def index)

Either ignore these or turn them off.  We generally use CodeClimate to measure complexity, but we need to make upgrades to the way CodeClimate is hooked up.

app/assets/stylesheets/partials/anime-page.css.scss
Eschon: Well I use Atom and as far as I know it adds an empty line at the end of the file by default.
What really confuses me is that as far as I understand it this diff says that the line was removed but it is there for me.

Anyway I'll just checkout another branch take the unchanged file from there and replace it. Don't know why I didn't think of that in the first place.

app/controllers/users_controller.rb
Eschon: I'll just try changing the order and see how it turns out, thanks.

Eschon:
> Either ignore these or turn them off. We generally use CodeClimate to measure complexity, but we need to make upgrades to the way CodeClimate is hooked up.

I'll ignore them for now but if you want me to turn them of in the .rubocop.yml I can do that to.

Eschon:
Another Question:
If I do further changes should I squash all my commit or is it ok to just add commits?

app/controllers/users_controller.rb
```diff
@@ -138,17 +139,20 @@ def update
...
       changes.delete(key)
     end
 
-    changes = changes.permit(:about, :location, :waifu, :website, :waifu_or_husbando,
-                   :bio, :waifu_char_id, :email, :cover_image, :sfw_filter,
-                   :title_language_preference, :password, :name, :star_rating)
+    changes = changes.permit(:about, :location, :website, :name, :waifu_char_id,
+                             :sfw_filter, :waifu, :bio, :email, :cover_image,
+                             :waifu_or_husbando, :title_language_preference,
+                             :password, :star_rating)
```
Eschon: Looks much better to me now.
What do you think?

wopian:
> @Eschon If I do further changes should I squash all my commit or is it ok to just add commits?

Squash all the commits once you're done.

Eschon:
Ok. Are there things that I should still change here otherwise I'll squash everything.

app/controllers/users_controller.rb
```diff
@@ -116,8 +117,8 @@ def comment
 
     # Create the story.
     @user = User.find(params[:user_id])
-    story = Action.broadcast(
-      action_type: "created_profile_comment",
+    _story = Action.broadcast(
```
NuckChorris: I assume `_story` is not used anywhere, hence the underscore — probably cleaner to just remove the assignment

NuckChorris:
Looks great to me.  I don't see anything besides the `_story` :+1: It's nice to finally have it clean lol

Eschon:
Ok everything should be ready now.

vevix:
:+1: awesome!

PR #563, opened 2015-05-27
Fix MAL Importer Prefixing "Edit" To Synopsis

wopian:
MAL recently changed the anime page to include links that were previously hidden if you were logged out - thus adding a `div` containing `Edit` which doesn't get stripped out by Hummingbird.

Fixes this:


DNA:
:shipit:

vevix:
:+1:

PR #564, opened 2015-05-27
Added padding to normal layout to align studio and genre lists when s…

### LINKED ISSUE #548: Studios not correct aligned when they enter a third line
unbenannt

As seen in the screenshot the alignment breaks when entering the third line. This would probably also be the case for genres but I have found no anime where this is the case.

astraldragon:
Fixes #548 
Added padding to normal layout to align studio and genre lists when spanning 3 or more lines


NuckChorris:
:dancer: LGTM!

PR #566, opened 2015-06-02
Rake task for anime images download

DNA:
With the current DB dump used for dev install, we don't got the anime images needed to correctly display the lists and individual anime pages (well, quite obvious, but still…), so I created a simple task to handle this:

```
rake hummingbird:images                    # Download images from Hummingbird for better development
rake hummingbird:images:covers[quantity]   # Download only anime covers
rake hummingbird:images:posters[quantity]  # Download only anime posters
```

By default it get images from 2 pages from default listing. If more (or less) images are needed, you can specify the value on each individual task.

If needed, we can add more complex tasks in the future (get images by genre, date, by individual id, casts, etc)

### DIFF
--- a/lib/tasks/hb.rake
+++ b/lib/tasks/hb.rake
@@ -28,18 +28,12 @@ namespace :hummingbird do
         puts "#{anime.title} - \033[31mNo image found, downloading…\033[0m"
 
         remote_url = open(
-          "https://hummingbird.me/full_anime/#{anime.slug}.json",
-          ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE
+          "https://hummingbird.me/full_anime/#{anime.slug}.json"
         ).read
         remote_anime = JSON.parse(remote_url)['full_anime']
 
-        anime.update_from_pending(
-          anime.create_pending(
-            User.first,
-            type => remote_anime[type.to_s],
-            edit_comment: 'Rake automatic image download'
-          )
-        )
+        anime.update_attributes(type => remote_anime[type.to_s])
+
       end
     end
   end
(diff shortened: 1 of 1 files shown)


lib/tasks/hb.rake
```diff
@@ -0,0 +1,46 @@
...
+        if path && File.exist?(path)
+          puts "#{anime.title} - \033[32mImage already downloaded\033[0m"
+          next
+        end
+
+        puts "#{anime.title} - \033[31mNo image found, downloading…\033[0m"
+
+        remote_url = open(
+          "https://hummingbird.me/full_anime/#{anime.slug}.json",
+          ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE
```
NuckChorris: Is there a reason you disable ssl verification?  Our SSL should be valid, and it's generally a bad thing to disable verification.  Definitely not a practice which is encouraged.

lib/tasks/hb.rake
```diff
@@ -0,0 +1,46 @@
...
+        ).read
+        remote_anime = JSON.parse(remote_url)['full_anime']
+
+        anime.update_from_pending(
```
NuckChorris: I wouldn't even bother going through the wiki system, honestly.  It's probably easier and faster to just update anime directly

lib/tasks/hb.rake
DNA: I fucked up all my ssl certificates and forgot to remove it from the code. Works nice without it on docker, tho

lib/tasks/hb.rake
DNA: Ok, I'll skip this part

DNA:
Ok, @NuckChorris,we're all set to go! :P

NuckChorris:
:+1: LGTM!

PR #569, opened 2015-06-03
fix link to users profile

vevix:
Issue with providing `currentUser`.

It redirects to the profile fine, but sets the URL `/users/undefined` :'(.


NuckChorris:
Seriously I thought Ember handled this but I guess not

NuckChorris:
Also LGTM

Issue #416, opened 2015-02-18
Show More on followed stories does not display all users


UberMouse:
It only expands to show another 5 rows and there is no scrollbar to view the rest

![Bug](http://i.imgur.com/mskECyz.png)

NuckChorris:
I think I see a few pixels of the next row at the bottom.  Perhaps this is an issue with overflow in CSS?

Wopian:
>It only expands to show another 5 rows and there is no scrollbar to view the rest

All 65 follows are listed.

> Perhaps this is an issue with overflow in CSS?

There's a `500px` max height on `.comment-text`.

vevix:
Can we get a PR on this? :open_mouth:

Wopian:
:fu: :no_good:

AinaM:
If we change the overflow or max-height for `.comment-text`, it will fix the problem here but I think there will be an issue will all other contents. Because the class `.comment-text` is used at several places.

__Exemple with a simple post content__
Normal : 

![normal_story](https://cloud.githubusercontent.com/assets/5560784/6749160/bc60bb20-ceed-11e4-90f1-8f6e9b72f263.png)

If we add `overflow: auto` on `.comment-text` :

![with oveflow](https://cloud.githubusercontent.com/assets/5560784/6749186/0174d4bc-ceee-11e4-8433-b8e90e7285ce.png)
It also changes here :worried:

Actually, I suppose that when clicking on "_Show more_" / "_Show less_", we should toggle `.full-post` on `.comment-text`. As [here](https://github.com/hummingbird-me/hummingbird/blob/4c7dad8683ddac2ca32c698b767044ba5d9b1cb8/frontend/app/templates/story/comment.hbs#L47) for post content :
``` handlebars
<p {{bind-attr class=":comment-text isExpanded:full-post" }}>
```
This class removes the max-height for `.comment-text`  ! :smile: 

Or, perhaps we could add [here](https://github.com/hummingbird-me/hummingbird/blob/4c7dad8683ddac2ca32c698b767044ba5d9b1cb8/frontend/app/templates/story/followed.hbs#L5) a specific class.
Exemple in template :
``` handlebars
<div class="comment-content comment-followed">
```
In scss : 
``` scss
.comment-followed {
  overflow: auto;
}
```

PR #570, opened 2015-06-05
Fix overflow issue with show more on followed stories

AinaM:
In reference to issue #416 :)

Here is how it looks like after clicking on "Show more", there is now a scrollbar to see everything :

screen_follow-show-more

### DIFF
--- a/app/assets/stylesheets/partials/status-panel.css.scss
+++ b/app/assets/stylesheets/partials/status-panel.css.scss
@@ -194,11 +194,7 @@
     &.full-post {
       max-height: none;
     }
-      
-    .follow-bubbles {
-      max-height: 480px;
-      overflow: auto;
-    }
+
   }
 
   .event-info {
--- a/app/assets/stylesheets/partials/status-panel.css.scss
+++ b/app/assets/stylesheets/partials/status-panel.css.scss
@@ -194,6 +194,11 @@
     &.full-post {
       max-height: none;
     }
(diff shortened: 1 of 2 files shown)


wopian:
Should this not be in `user-activity-feed.css.scss` which has this styling class already? 

```css
.follow-bubbles {
  padding-left: 15px;
  padding-right: 15px;
  margin-top: 15px;
  ...
}
```

https://github.com/hummingbird-me/hummingbird/blob/8e30fd95d6438adb8783c4ab6d16245896a75490/app/assets/stylesheets/partials/user-activity-feed.css.scss#L452

AinaM:
I hesitated, but if we want to use `.follow-bubbles` somewhere else, it may not be pratical :cry: . 
Another thing is that the value 480px is linked to the 500px `max-height` on `.comment-text`. If the `max-height` of `.comment-text` changes, it will be easier to remember to change the `max-height` for `.follow-bubbles` if it is in the same file too, no ?
If I'm wrong, I can change :smile:

JoshFabian:
@AinaM sorry for the delay on getting this reviewed. Keeping the css neater where we can is preferred, follow bubbles likely won't be going beyond the feed. The max-height of 480 here makes sense to me so I'd just move the scss to the correct place and this is good to go. Also, let's get you into the slack dev chat! Shoot me an email for an invite josh@hummingbird.me :)

PR #576, opened 2015-06-07
Rubocop

Eschon:
I took on the next file. Again like last time i ignored the ABC-size and method has too many lines errors.

But again there were two errors where I didn't know how to handle:
 * Use a guard clause instead of wrapping the code inside a conditional expression (I'll add a comment for the line)
 * Don't use `time.now` without .zone. Can I just change it to `time.zone.now` without any side effects?

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,53 +1,54 @@
 require_dependency 'mal_import'
 
 class AdminController < ApplicationController
-  before_filter :allow_only_admins
+  before_action :allow_only_admins
 
   def allow_only_admins
     # This shouldn't be needed becuse we also check for admin-ness in the
     # routes. Still doing this just to be safe.
     authenticate_user!
-    if not current_user.admin?
-      raise ActionController::RoutingError.new('Not Found')
-    end
+
+    fail ActionController::RoutingError, 'Not Found' unless current_user.admin?
   end
 
   def find_or_create_by_mal
     media = params[:media].to_sym
     mal = MALImport.new(media, params[:mal_id].to_i, :shallow)
     @thing = case media
-    when :anime
-      Anime.create_or_update_from_hash mal.to_h
-    when :manga
-      Manga.create_or_update_from_hash mal.to_h
+             when :anime
+               Anime.create_or_update_from_hash mal.to_h
+             when :manga
+               Manga.create_or_update_from_hash mal.to_h
     end
     redirect_to @thing
     MyAnimeListScrapeWorker.perform_async(media, params[:mal_id].to_i)
   end
 
   def index
-    @anime_without_mal_id = Anime.where(mal_id: nil).where(%{"anime"."id" NOT IN (SELECT "anime_genres"."anime_id" FROM "anime_genres" INNER JOIN "genres" ON "genres"."id" = "anime_genres"."genre_id" AND "genres"."name" = 'Anime Influenced')})
+    @anime_without_mal_id = Anime.where(mal_id: nil).where.not(
+      id: Anime.select(:id)
+               .joins(:genres)
+               .where(genres: { name: 'Anime Influenced' })
+    )
     # Sort by partner code count (ascending) where there's less than 20 left
-    @deals = PartnerDeal.joins('left join partner_codes on partner_codes.partner_deal_id = partner_deals.id').
-      select('partner_deals.*, count(partner_codes.id) as codes_remaining').
-      group('partner_deals.id').
-      having('count(partner_codes.id) < 20').
-      order('count(partner_codes.id) asc')
+    @deals = PartnerDeal
+             .joins(
+               'LEFT JOIN partner_codes \
+                ON partner_codes.partner_deal_id = partner_deals.id'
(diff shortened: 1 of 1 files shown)


app/controllers/admin_controller.rb
```diff
@@ -7,32 +7,42 @@ def allow_only_admins
...
+             )
+             .select(
+               'partner_deals.*, count(partner_codes.id) AS codes_remaining'
+             )
+             .group('partner_deals.id')
+             .having('count(partner_codes.id) < 20')
+             .order('count(partner_codes.id) asc')
     @blotter = Blotter.get
 
     if params[:old_kotodama].nil?
```
Eschon: Here it displays the guard clause error. I guess putting everything inside the if won't work. What should I do about it?

NuckChorris: What it's basically saying is "don't just wrap everything in an if statement when you mean to prevent it from running in a certain case"

In this case, though, the answer is probably removing the entire conditional — `old_kotodama` isn't used anymore (or ever, I think)

NuckChorris:
> Don't use time.now without .zone. Can I just change it to time.zone.now without any side effects?

Basically, `.zone` is to expose the data in the user's timezone.  This isn't really relevant since we use Ember — all Rails deals with is UTC.  If you can find the Rubocop for that, nuke it.

app/controllers/admin_controller.rb
```diff
@@ -7,32 +7,42 @@ def allow_only_admins
     # This shouldn't be needed becuse we also check for admin-ness in the
     # routes. Still doing this just to be safe.
     authenticate_user!
-    if not current_user.admin?
-      raise ActionController::RoutingError.new('Not Found')
-    end
+
+    fail ActionController::RoutingError 'Not Found' unless current_user.admin?
```
DNA: Tests are throwing and error here:

`Message: <"undefined method 'RoutingError' for ActionController:Module">`

without the `new()` method, ruby think that `RoutingError` is a method and not a class

NuckChorris: It should be `fail ActionController::RoutingError, 'Not Found'` (note the comma) — Ruby is thinking that there's implicit parentheses like this, that it's `fail(ActionController::RoutingError('not found'))`

NuckChorris:
If you rebase this on top of `master`, it'll probably complain a lot less.  @DNA made some awesome changes to the `.rubocop.yml` to remove most irrelevant things.

Eschon:
Sorry for the wait, I was at a music festival last weekend and needed a little time to recover ^^

I made the two changes that @NuckChorris and @DNA pointed out.
I also rebased like @NuckChorris suggested and now all remaining offenses on this file are gone.

When you still want things changed here tell me otherwise I can squash everything.

app/controllers/admin_controller.rb
```diff
@@ -7,47 +7,49 @@ def allow_only_admins
...
     MyAnimeListScrapeWorker.perform_async(media, params[:mal_id].to_i)
   end
 
   def index
-    @anime_without_mal_id = Anime.where(mal_id: nil).where(%{"anime"."id" NOT IN (SELECT "anime_genres"."anime_id" FROM "anime_genres" INNER JOIN "genres" ON "genres"."id" = "anime_genres"."genre_id" AND "genres"."name" = 'Anime Influenced')})
+    @anime_without_mal_id = Anime.where(mal_id: nil).where(
+      %{"anime"."id" NOT IN (SELECT "anime_genres"."anime_id" \
+         FROM "anime_genres" INNER JOIN "genres" \
+         ON "genres"."id" = "anime_genres"."genre_id" \
+         AND "genres"."name" = 'Anime Influenced')}
```
NuckChorris: Heredoc syntax is often more legible for these fatass SQL statements:

```ruby
@anime_without_mal_id = Anime.where(mal_id.nil).where(<<-SQL
  "anime"."id" NOT IN (
    SELECT "anime_genres"."anime_id"
    FROM "anime_genres" INNER JOIN "genres"
    ON "genres"."id" = "anime_genres"."genre_id"
    AND "genres"."name" = 'Anime Influenced'
  )
SQL)
```

DNA: We can let ActiveRecord handle it:
```ruby
@anime_without_mal_id = Anime.where(mal_id: nil).where.not(
  id: Anime.select(:id)
           .joins(:genres)
           .where(genres: { name: 'Anime Influenced' })
)
```

DNA:
I've tested it here and I got another offense:
```
app/controllers/admin_controller.rb:4:3: C: Prefer before_action over before_filter.
  before_filter :allow_only_admins
  ^^^^^^^^^^^^^
```
Both methods do the same thing, but if we plan to move to rails 4.2, I think it's good to change this too `before_action` too. After this and the `anime_without_mal_id` query we are ready to go :) :shipit:

Eschon:
Hmm, strange I don't get that offense, but I changed it anyways.
I also let Active record handle the fatass SQL statement and squashed everything.

NuckChorris:
:+!: LGBT

Issue #575, opened 2015-06-07
Anime genres are duplicated on Dashboard and User pages.


wopian:
![](http://i.imgur.com/b4gKy3u.png)

Although it seems *sometimes* it'll only show it once (maybe its because it only has 1 genre?)

![](http://a.pomf.se/snebqe.png)

Reported in this story: https://hummingbird.me/stories/5477299

vevix:
Seems to be caused by the addition of the `default_scope` on the anime model.

NuckChorris:
Oh god you're probably right. The story controller tries to do some funky shit with autoloading

DNA:
Well, at least the scope fix are already done on #577, we just need to send it to production :P 

(I laughed at "funky shit with autoloading" xD)

NuckChorris:
Sadly that scope fix probably won't fix this.  This is because we preload the genres in that scope (good!) but the `SearchQuery` forcibly preloads things via some strange hacks.  We probably need to `.unscoped` it

DNA:
~some strange hacks~ issue resolved xD

wopian:
Fixed.

PR #577, opened 2015-06-08
Default scopes are fucking evil

DNA:
I was certain that a huge PR like that would mess with something… Sorry guys! :disappointed:

### DIFF
--- a/lib/story_query.rb
+++ b/lib/story_query.rb
@@ -1,28 +1,30 @@
 require_dependency 'user_query'
 
 class StoryQuery
-
   # Returns a list of stories having the given IDs.
   def self.find_by_ids(story_ids, current_user)
-    stories = Story.where(id: story_ids).for_user(current_user)
-                   .includes(:user, :substories, :target,
-                             substories: [:user, :target])
+    stories = Story.where(id: story_ids)
+              .for_user(current_user)
+              .includes(:user, :substories, :target,
+                        substories: %i(user target))
 
     # Preload genres for anime stories.
-    anime_stories = stories.select {|s| s.target_type == "Anime" }.map(&:target)
-                   .index_by(&:id)
+    stories.select { |s| s.target_type == 'Anime' }.map(&:target).index_by(&:id)
 
-    comment_stories = stories.select {|x| x.story_type == "comment" }
+    comment_stories = stories.select { |x| x.story_type == 'comment' }
     comment_index = comment_stories.index_by(&:id)
 
     # Load is_liked for comment stories.
     if current_user
-      Vote.where(user_id: current_user.id, target_type: "Story",
-                 target_id: comment_index.keys).select(:target_id).each do |vote|
+      Vote.where(
+        user_id: current_user.id,
+        target_type: 'Story',
+        target_id: comment_index.keys
+      ).select(:target_id).each do |vote|
         comment_index[vote.target_id].set_is_liked! true
       end
     end
-    comment_stories.select {|x| x.is_liked.nil? }.each do |story|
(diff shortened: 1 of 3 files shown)


NuckChorris:
Are there any places you need to re-add the order?  Otherwise :+1: LGTM

DNA:
No there isn't. I added it to have a "default order" instead of db natural order. It obviously didn't work as intended… xD

NuckChorris:
Remember there's no such thing as "natural order" only "undefined and potentially random order" — anything which uses offsets or pagination needs an explicit order

DNA:
Added a commit to also fix #575. Also rubocop'd the StoryQuery

lib/story_query.rb
```diff
@@ -1,34 +1,30 @@
...
     # Preload genres for anime stories.
-    anime_stories = stories.select {|s| s.target_type == "Anime" }.map(&:target)
-                   .index_by(&:id)
-    Anime.where(id: anime_stories.values.map(&:id)).includes(:genres).each do |a|
-      assoc = anime_stories[a.id].association(:genres)
-      assoc.loaded!
-      assoc.target.concat(a.genres)
-      a.genres.each {|genre| assoc.set_inverse_instance(genre) }
-    end
+    stories.select { |s| s.target_type == 'Anime' }.map(&:target).index_by(&:id)
```
NuckChorris: We shouldn't remove this unless we know for certain that it won't introduce an N+1 query.  I'm not sure the eager loading in the default scope will save us here with nested associations (stories -> anime -> genres) especially when they're both polymorphic...

DNA: There wasn't any N+1 problems when I tested it, but I'll map the `StoryQuery` usage and check for potential problems.

DNA: The loading seems fine:
```
Story Load (1.2ms)  SELECT  "stories"."id" FROM "stories" LEFT OUTER JOIN watchlists ON watchlists.id = stories.watchlist_id WHERE "stories"."deleted_at" IS NULL AND "stories"."user_id" = $1 AND (NOT adult) AND (watchlists.id IS NULL OR watchlists.private = 'f') AND (stories.group_id IS NULL)  ORDER BY stories.updated_at DESC LIMIT 30 OFFSET 0  [["user_id", 1]]
Story Load (2.7ms)  SELECT "stories".* FROM "stories" LEFT OUTER JOIN watchlists ON watchlists.id = stories.watchlist_id WHERE "stories"."deleted_at" IS NULL AND "stories"."id" IN (1, 2) AND (NOT adult) AND (watchlists.id IS NULL OR watchlists.private = 'f')
Substory Load (2.4ms)  SELECT "substories".* FROM "substories"  WHERE "substories"."deleted_at" IS NULL AND "substories"."story_id" IN (2, 1)
Anime Load (1.7ms)  SELECT "anime".* FROM "anime"  WHERE "anime"."id" IN (3936)
HABTM_Genres Load (1.8ms)  SELECT "anime_genres".* FROM "anime_genres"  WHERE "anime_genres"."anime_id" IN (3936)
Genre Load (2.8ms)  SELECT "genres".* FROM "genres"  WHERE "genres"."id" IN (28, 8, 11, 4, 2, 1)
```
I've added all the possibilities I know that would break things, and everything seems ok. If you remember some edge case, lmk

NuckChorris:
:+1: LGTM (if it causes regressions we may have to back it out though)

DNA:
I broke user stories once, I don want to do it twice… xD

NuckChorris:
Deployed! Looks like nothing broke :tanabata_tree: :8ball:

PR #578, opened 2015-06-08
Create Media concern to DRY Anime and Manga models

DNA:
The title is pretty straight forward. My first thought was to mode paperclip related methods inside a concern, but I saw many other methods that we could DRY, so I made a concern grouping all duplicated code. 

I think the concern has the right size, but if you guys like we can split this concern into Media and MediaImage concerns

### DIFF
--- a/app/models/manga.rb
+++ b/app/models/manga.rb
@@ -31,80 +31,21 @@
 #
 
 class Manga < ActiveRecord::Base
-  include Versionable
-  include BayesianAverageable
-
-  include PgSearch
-  pg_search_scope :instant_search,
-    against: [ :romaji_title, :english_title ],
-    using: { tsearch: { normalization: 42, dictionary: 'english' } },
-    ranked_by: ':tsearch'
-  pg_search_scope :full_search,
-    against: [ :romaji_title, :english_title ],
-    using: {
-      tsearch: { normalization: 42, dictionary: 'english' },
-      trigram: { threshold: 0.1 }
-    },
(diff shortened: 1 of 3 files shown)


DNA:
Important: This PR also replaces #530 and fix #512

NuckChorris:
Looks like you need to rebase

DNA:
Rebased! :shipit:

app/models/manga.rb
```diff
@@ -31,82 +31,23 @@
...
-    against: [ :romaji_title, :english_title ],
-    using: {
-      tsearch: { normalization: 42, dictionary: 'english' },
-      trigram: { threshold: 0.1 }
-    },
-    ranked_by: ':tsearch + :trigram'
-
-  extend FriendlyId
-  friendly_id :romaji_title, use: [:slugged, :history]
+  VALID_TYPES =  ["Manga", "Novel", "One Shot", "Doujin", "Manhwa", "Manhua", "OEL"]
```
NuckChorris: I went over this PR like 4 times trying to find any problems (was mostly thinking with `self`, since we had a couple of those mess up in a previous refactor) but the only thing I was able to come up with was that these should be single quotes since they don't have interpolation.  Everything else looks fine.

Fix 'em and LGTM :+1:

DNA:
Class methods are easy to spot on concerns since it needs to be inside `ClassMethods`. But that's not what we need, so LGTM :+1:  :shipit:

PR #579, opened 2015-06-08
Rubocop anime controller

DNA:
This PR add some changes to rubocop rules and refactor the Anime Controller. 

Rubocop rules removed:
- **Rails/HasAndBelongsToMany** because we use habtm associations a lot
- **Rails/TimeZone** to stop receiving errors about time.zone, since we are all on 
- Some rules were removed because we use Code Climate to handle that:
  - **Metrics/CyclomaticComplexity**
  - **Metrics/PerceivedComplexity**
  - **Metrics/AbcSize**
  - **Metrics/MethodLength**

### DIFF
--- a/app/controllers/anime_controller.rb
+++ b/app/controllers/anime_controller.rb
@@ -3,12 +3,14 @@ require 'wilson_score'
 class AnimeController < ApplicationController
   include MicrodataHelper
   def random
-    anime = Anime.where("age_rating <> 'R18+'").where(show_type: ["TV", "OVA", "ONA", "Movie"]).order("RANDOM()").limit(1)[0]
-    redirect_to anime
+    redirect_to Anime.sfw_filter(current_user)
+                .where(show_type: %w(TV OVA ONA Movie))
+                .order('RANDOM()')
+                .first
   end
 
   def show
-    @anime = Anime.includes(:genres).find(params[:id])
+    @anime = Anime.find(params[:id])
 
     respond_to do |format|
       format.html do
@@ -17,7 +19,9 @@ class AnimeController < ApplicationController
           return redirect_to @anime, status: :moved_permanently
         end
 
-        preload_to_ember! @anime, serializer: FullAnimeSerializer, root: "full_anime"
+        preload_to_ember! @anime,
+                          serializer: FullAnimeSerializer,
+                          root: 'full_anime'
         render layout: 'redesign'
       end
       format.json do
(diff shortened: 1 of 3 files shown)


app/controllers/anime_controller.rb
```diff
@@ -3,12 +3,14 @@
 class AnimeController < ApplicationController
   include MicrodataHelper
   def random
-    anime = Anime.where("age_rating <> 'R18+'").where(show_type: ["TV", "OVA", "ONA", "Movie"]).order("RANDOM()").limit(1)[0]
-    redirect_to anime
+    redirect_to Anime.sfw_filter(current_user)
+      .where(show_type: %w(TV OVA ONA Movie))
```
NuckChorris: Doesn't rubocop nag you to align the dots?

app/controllers/anime_controller.rb
```diff
@@ -185,13 +203,12 @@ def filter_new
...
 
-        if params[:sort]
-          @sort = params[:sort]
-        else
-          @sort = "all"
-        end
+        @sort = params[:sort] ||= 'all'
 
-        @anime = Anime.accessible_by(current_ability).references(:genres).page(params[:page]).per(36)
+        @anime = Anime.accessible_by(current_ability)
```
NuckChorris: We can probably remove the `accessible_by` here, since we're no longer using Cancan

app/controllers/anime_controller.rb
```diff
@@ -201,35 +218,39 @@ def filter_new
...
+            condition = arel[:started_airing_date]
+                        .in(range)
+                        .or(arel[:finished_airing_date].in(range))
             sum.class == Arel::Table ? condition : sum.or(condition)
           end
-          if @years.include? "Upcoming"
-            condition = arel[:started_airing_date].gt(Time.now.to_date).or(arel[:started_airing_date].eq(nil))
+          if @years.include? 'Upcoming'
+            condition = arel[:started_airing_date]
+                        .gt(Time.zone.now.to_date)
```
NuckChorris: <strike>`.zone`</strike>

app/controllers/anime_controller.rb
DNA: Looking closely, I think its aligning the same way it would to `redirect_to` arguments, instead of `Anime` methods. A bug, maybe?

NuckChorris: Perhaps.  Either way it should probably be aligned (humans matter more than rubocop)

DNA:
Done!

app/controllers/anime_controller.rb
```diff
@@ -95,13 +111,9 @@ def filter_old
       @years = @filter_years
     end
 
-    if params[:sort]
-      @sort = params[:sort]
-    else
-      @sort = "all"
-    end
+    @sort = params[:sort] ||= 'all'
```
NuckChorris: Since you don't need to assign params[:sort] you can just do `params[:sort] || 'all'`

app/controllers/anime_controller.rb
```diff
@@ -111,35 +123,39 @@ def filter_old
...
     # Apply sort option.
-    if @sort == "newest"
-      @anime = @anime.order("started_airing_date DESC")
-    elsif @sort == "oldest"
-      @anime = @anime.order("started_airing_date")
-    elsif @sort == "popular"
-      @anime = @anime.order("user_count DESC NULLS LAST")
+    if @sort == 'newest'
+      @anime = @anime.order('started_airing_date DESC')
+    elsif @sort == ''
```
NuckChorris: There a reason you emptied the `'oldest'` string?

DNA: It sure was by mistake. I'll add it back

DNA:
'oldest' string fixed and commits squashed :)

NuckChorris:
LGTM

PR #583, opened 2015-06-13
Oauth2 provider

NuckChorris:
So, this is a work in progress, but it's already at <strike>22</strike> <strike>27</strike> <strike>31</strike> <strike>36</strike> <strike>38</strike> <i>42</i> files and there's a lot of security-related code.  So I figure we should start review early.  Because I expect probably <strike>10 more files will be added.</strike> 15 files later, and still nowhere near complete.  Should hit 40 files at least.

 * [x] Update apps management panel to expose all fields
 * [x] Add new Token types and have a class method on Token which figures out which to use
 * [x] Add Authorization endpoint
 * [x] Add Token endpoint
    * [x] `authorization_code`
    * [x] `refresh_token`
    * [x] `password`
 * [x] Store authorizations in a database
 * [x] Redirect to login/signup page if not logged in
   * [x] Add parameter to `sign-up` and `sign-in` routes in Ember to provide a post-sign-in redirect target
      * [x] Maintain that parameter when switching between the two
 * [x] `state` parameter

I was going to include the User endpoint (first endpoint to make use of the OAuth2) in this PR but it's already pretty big.

Technically we can merge this as-is already, if we decide we want to split the actual OAuth2 endpoints into a separate PR.  This is almost entirely just refactoring.


NuckChorris:
All checks done, all tests passing; all we need now are the pages.  We're ready for review.

I'm gonna set up a public version of this on my server and start aiming OAuth2 libraries at it to make any adjustments for compatibility.  That said, I doubt we'll have _too much_ trouble.

app/controllers/oauth2/authorization_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+# Throughout this file there are references to RFC 6749 in parentheses.  The
+# numbers separated by dots are the section id from the spec, the number after
+# "P" is the paragraph, and "S" is sentence.
+class OAuth2::AuthorizationController < ApplicationController
+  # Valid response errors (4.1.2.1, 4.2.2.1)
+  VALID_ERRORS = %i[invalid_request unauthorized_client access_denied
+                    unsupported_response_type invalid_scope server_error
+                    temporarily_unavailable]
+  # Not specified by RFC 6749, these are internal
+  VALID_FATAL_ERRORS = %i[invalid_client invalid_redirect]
```
DNA: Dunno if it will impact the implementation rules, but wouldn't be better to move those validations to a lib and throw exceptions that we can handle in the controller? It seems to be getting a little fat with all these private methods and all those RFC comments

app/controllers/oauth2/authorization_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+  # Redirect back to the application with payload
+  def redirect_back!(obj)
+    # State parameter to help clients prevent CSRF (10.12)
+    obj = { state: params[:state] }.merge(obj).compact
+
+    uri = if params[:response_type] == 'token'
+            client.build_redirect_uri(:fragment, obj)
+          else
+            client.build_redirect_uri(:query_string, obj)
+          end
```
DNA: Won't be better to read to use a ternary rather than an if on the assignment?
```ruby
type = params[:response_type] == 'token' ? :fragment : :query_string
uri = client.build_redirect_uri(type, obj)

File: app/controllers/oauth2/authorization_controller.rb [hunk 3]
```diff
@@ -0,0 +1,150 @@
...
+
+  def redirect_back_with_token!
+    if params[:response_type] == 'token'
+      token = OAuth2::Token.new(current_user, client, scopes)
+      redirect_back! token: token
+    else
+      code = OAuth2::Code.new(current_user, client, scopes)
+      redirect_back! code: code
+    end
+  end
```
DNA:
```ruby
if params[:response_type] == 'token'
  type = { token: OAuth2::Token.new(current_user, client, scopes) }
else
  type = { code: OAuth2::Code.new(current_user, client, scopes) }
end
redirect_back! type

File: app/controllers/oauth2/token_controller.rb [hunk 4]
```diff
@@ -0,0 +1,131 @@
...
+  GRANT_TYPES = %w[authorization_code refresh_token password client_credentials]
+  # Valid response errors (5.2)
+  VALID_ERRORS = {
+    invalid_request: 400,
+    invalid_client: 401,
+    invalid_grant: 400,
+    invalid_scope: 400,
+    unauthorized_client: 400,
+    unsupported_grant_type: 400,
+    server_error: 500
```
DNA: The comment on application_controller.rb is also valid here

app/controllers/oauth2/authorization_controller.rb
NuckChorris: Ah you're right

app/controllers/oauth2/authorization_controller.rb
NuckChorris: Many of the private methods are just shortcuts for things in params or headers, so those can't be moved.  The validations are also (a) too complex to express via `ActiveModel::Validators` and (b) the entirety of this controller (literally it's just validating and one line of "oh right we should go render that page"

Due to the lack of overlap in implementation between this and the `OAuth2::TokenController`, I don't think it's worthwhile right now to refactor.

DNA: Point taken :)

app/controllers/oauth2/token_controller.rb
```diff
@@ -0,0 +1,131 @@
...
+    if has_basic_credentials?
+      basic_credentials
+    elsif params.key?(:client_id) && params.key(:client_secret)
+      [params[:client_id], params[:client_secret]]
+    else
+      nil
+    end
+  end
+
+  def error!(err, message = nil, uri: nil)
```
DNA: Is there a reason to use a named argument on `uri` and not on `message`?

NuckChorris: Not really, I think I originally just had message and then added uri later

app/models/oauth2/code.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# This can be traded by the bearer of the client_secret for an actual Token
+# with the desired scopes.  Basically just an IOU for a Token.
+class OAuth2::Code < OAuth2::Token
+  def initialize(user, client, scopes, opts={})
+    opts = {
+      exp: (Time.now + 5.minutes.to_i).to_i,
+      token_scopes: scopes
+    }.merge(opts)
+    super(user, client, %i[oauth2_code], opts)
```
DNA: since it's only one symbol, wouldn't be better to just pass `:oauth2_code`?

app/models/oauth2/refresh_token.rb
```diff
@@ -0,0 +1,19 @@
...
+# This can be traded by the bearer of the client_secret for a new Token
+# with the desired scopes.  Basically just an IOU for a Token, which is rarely
+# transmitted over the network and thus remains more secure in the long-term.
+class OAuth2::RefreshToken < OAuth2::Token
+  def initialize(user, client, scopes, opts={})
+    opts = {
+      exp: (Time.now + 6.months.to_i).to_i,
+      token_scopes: scopes
+    }.merge(opts)
+    super(user, client, %i[oauth2_refresh], opts)
```
DNA: Same here

app/models/user.rb
```diff
@@ -83,6 +83,12 @@ def self.find_by_username(username)
     where('LOWER(name) = ?', username.to_s.downcase).first
   end
 
+  def self.find_by_login(name_or_email)
+    key = name_or_email.include?('@') ? :email : :name
+
+    User.where("lower(#{key}) = ?", name_or_email.downcase).first
```
DNA: why use both `lower()` and `.downcase`??

app/models/oauth2/code.rb
NuckChorris: No it's an array

DNA: `[:oauth2_code]` then? :P

DNA:
Phew! That was a long PR to review… There was just some code style and such that I've found, but haven't found any problem. It'll be a nice feature! :)

PR #585, opened 2015-06-14
Use ||= for constant declarations

NuckChorris:
This prevents "already initialized constant" warnings when Rails reloads the class.  Sadly there doesn't seem to be anything in rubocop to make this reminder automatic.

We should send this upstream to rubocop and take note of this in future PRs.  Also it looks like this might not apply to things inheriting from Rails-provided objects.


DNA:
:+1:

PR #586, opened 2015-06-14
Fix paperclip thumb generation

DNA:
`poster_image_thumb` was a method used to solve a problem on paperclip gem. My last attempt to fix it fixed small images, but added a bug on images with `.jpeg` extension.

Talking with @NuckChorris we decided that was better to monkeypatch the gem rather than trying to create a method to cover all images possibilities.


NuckChorris:
:+1: :fire: :100: LGTM!

PR #588, opened 2015-06-19
Updated gemfile - upgrade to rails 4.2.2

jacemonje:


NuckChorris:
We can't just update to 4.2, there's work which needs doing. See https://www.pivotaltracker.com/story/show/89862350

NuckChorris:
Additionally this would also need Gemfile.lock to be updated with `bundle update rails`

jacemonje:
I see. I didn't see that pivotal tracker ticket, sorry!

DNA:
For what I saw, we need to update the `devise` gem, add maybe add the `responder` gem, since devise need it in some cases (the docs says so, but they don't add it as a dependence). The `non-stupid-digest-assets` gem also seems to be fine.

But I'm pretty sure that all this changes will throw a lot of errors on the tests :P

NuckChorris:
Honestly I don't think it'll break too many tests once the libraries are updated and tweaked.

PR #589, opened 2015-06-22
Enable mini-profiler in production for admins

NuckChorris:
No better place to profile than production.

Eventually we should integrate this with a "ninja goggles mode" we talked about, perhaps just a link in the user menu which says "put on moderator pants"


vevix:
:+1:

NuckChorris:
Side thought: this is still hidden by CSS on Ember pages.  We should fix that.

PR #590, opened 2015-06-28
Fix ninja bans to apply everywhere

NuckChorris:
As it turns out, ninja bans didn't apply _anywhere_.  Like, literally nowhere.  Whoops.

I fixed it.


vevix:
looks good :+1:

DNA:
Seems fine, but it would be nice if it had some tests :P

PR #591, opened 2015-06-28
Resolve rubocop offenses for application_controller.rb

Eschon:
I fixed the offenses for another file.
As always I added line comments for the parts where I'm unsure if I have done it right.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from 'Auth::UnauthorizedException' do
     respond_to do |format|
-      format.json { render json: {error: "Not authenticated"}, status: 403 }
+      format.json { render json: { error: 'Not authenticated' }, status: 403 }
       format.html { redirect_to(new_user_session_path) }
     end
   end
@@ -25,13 +25,13 @@ class ApplicationController < ActionController::Base
 
   # Send an object along with the initial HTML response that will be loaded into
   # Ember Data's cache.
-  def preload_to_ember!(data, options={})
+  def preload_to_ember!(data, options = {})
     @preload ||= []
     data = ed_serialize(data, options)
     @preload << data unless data.nil?
   end
 
-  def ed_serialize(data, options={})
+  def ed_serialize(data, options = {})
     data = data.to_a if data.respond_to? :to_ary
     data = [data] unless data.is_a? Array
     return if data.empty?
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -104,18 +104,19 @@ def save_and_render(model)
...
 
   # DEPRECATED
-  def hide_cover_image;end
+  def hide_cover_image; end
 
   def mixpanel
     if Rails.env.production?
-      @mixpanel ||= Mixpanel::Tracker.new '92b66301c752642b40ca39e718517d94', { :async => true, :env => request.env }
+      @mixpanel ||= Mixpanel::Tracker.new '92b66301c752642b40ca39e718517d94',
+                                          async: true, env: request.env
```
Eschon: Here it said the { } where redundant so I removed them, but I'm not sure if that was right.

app/controllers/application_controller.rb
```diff
@@ -127,12 +128,12 @@ def after_sign_in_path_for(resource)
     stored_location_for(resource) || user_path(resource)
   end
 
-  def after_sign_out_path_for(resource)
+  def after_sign_out_path_for(_resource)
```
Eschon: For now I renamed resource, but if it is really not needed I can remove it.

app/controllers/application_controller.rb
DNA: It's ok. Since the first parameter is a string, and not a hash, ruby interpreter will split it into 2 method variables :)

app/controllers/application_controller.rb
NuckChorris: This might actually be unused now... I think it was something for the old Devise login pages

DNA: Yes, it can be removed :)

Eschon:
Done
I removed the unused argument and squashed everything.

app/controllers/application_controller.rb
```diff
@@ -127,12 +128,12 @@ def after_sign_in_path_for(resource)
     stored_location_for(resource) || user_path(resource)
   end
 
-  def after_sign_out_path_for(resource)
+  def after_sign_out_path_for
```
NuckChorris: If `after_sign_out_path_for` is still used anywhere (I doubt it), this will cause `ArgumentError: wrong number of arguments (1 for 0)`

Eschon:
I did like @NuckChorris suggested and removed `after_sign_out_path_for`.
Anything else that should be changed?

NuckChorris:
:+1: Looks great!

PR #594, opened 2015-07-07
Fix MAL Importer Importing Background Details

wopian:


lib/mal_import.rb
```diff
@@ -102,7 +102,7 @@ def metadata
         ja_jp: begin @sidebar.css('div:contains("Japanese:")')[0].text.gsub("Japanese: ", "") rescue nil end
       }.compact,
       synopsis: begin
-        synopsis = @main_noko.css('td td:contains("Synopsis")')[0].text.gsub("EditSynopsis", '').split("googletag.cmd.push")[0]
+        synopsis = @main_noko.css('td td:contains("Synopsis")')[0].text.gsub("EditSynopsis", '').split("EditBackground")[0].split("googletag.cmd.push")[0]
```
NuckChorris: Doesn't the `googletag.cmd.push` land in the part after `EditBackground`?

wopian: Left it in for backwards compatibility, In case they decide to move the background information into it's own table column.

NuckChorris: Well it doesn't hurt and makes things slightly more robust :+1:

PR #596, opened 2015-07-11
Confirm before story/substory deletion

NuckChorris:
A wretched and nasty hack, but effective nonetheless.


vevix:

PR #597, opened 2015-07-12
Cleanup indices

NuckChorris:
Tweaks to our indexing situation discovered via pghero and mini-profiler in production.

For the most part, these added indices shouldn't make a noticeable difference in load times.  `recommendations` may benefit, but `reviews` is such a small table that the current seq scan doesn't add more than about 15ms to the anime endpoint.

The `created_at` on `substories` should reduce the kotodama load time quite a bit (hoping to shave off about 5 seconds there)

There were also two completely nonsensical and pointless indices on the `users` table for `waifu_slug` and `waifu_or_husbando`.  We've never had any queries based on those, and I can't imagine ever needing them, so I removed them.

### DIFF
--- a/db/migrate/20150712193314_index_cleanup.rb
+++ b/db/migrate/20150712193314_index_cleanup.rb
@@ -1,0 +1,15 @@
+class IndexCleanup < ActiveRecord::Migration
+  def change
+    # Extremely common operations
+    add_index :reviews, :user_id
+    add_index :reviews, :anime_id
+    add_index :recommendations, :user_id
+
+    # WTF?
+    remove_index :users, :waifu_or_husbando
+    remove_index :users, :waifu_slug
+
+    # Used on Kotodama only, but avoids 6-second load times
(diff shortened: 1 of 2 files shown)


db/migrate/20150712193314_index_cleanup.rb
```diff
@@ -0,0 +1,15 @@
+class IndexCleanup < ActiveRecord::Migration
+  def change
+    # Extremely common operations
+    add_index :reviews, :user_id
+    add_index :reviews, :anime_id
+    add_index :recommendations, :user_id
+
+    # WTF?
+    remove_index :users, :waifu_or_husbando
```
DNA: Since there's a work on waifu/husbando on #375 (it's still active?) I think it's okay to remove this without the *wtf?* comment… :P

DNA:
:see_no_evil: :shipit:

vevix:
:heartbeat: :+1:

PR #598, opened 2015-07-12
Ignore Dropbox StorageQuota errors

NuckChorris:
Dropbox sends us errors for when a user is out of space.  Ideally, we should notify the user, but at the very least we should not retry it over and over.


vevix:
:+1:

PR #604, opened 2015-07-18
fix rubocop offenses for apps_controler.rb

Eschon:
Finally for the first time I have no questions after working on a file (probably because there wasn't much to fix :D)


NuckChorris:
No surprise there, it's a pretty recent addition. LGTM! Already!

PR #605, opened 2015-07-18
Resolve rubocop offenses for auth_controller.rb

Eschon:
Since the last one was merged so fast I decided to do another file.
This time I'm not sure about how I fixed the long lines. Does anyone have a better idea?

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -8,7 +8,9 @@ class AuthController < ApplicationController
   end
 
   def sign_in_action
-    return error! "Missing parameters", 400 unless [:email, :password].none? { |x| params[x].blank? }
+    unless %i(email password).none? { |x| params[x].blank? }
+      return error! 'Missing parameters', 400
+    end
 
     key = params[:email].include?('@') ? :email : :name
     user = User.where("lower(#{key}) = ?", params[:email].downcase).first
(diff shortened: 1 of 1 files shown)


app/controllers/auth_controller.rb
```diff
@@ -8,7 +8,10 @@ def sign_out_action
   end
 
   def sign_in_action
-    return error! "Missing parameters", 400 unless [:email, :password].none? { |x| params[x].blank? }
+    return error! 'Missing parameters', 400 unless %i(email password)
```
NuckChorris: Probably best to split it to a multi-line conditional

app/controllers/auth_controller.rb
```diff
@@ -17,20 +20,25 @@ def sign_in_action
...
 
   def sign_up_action
-    return error! "Missing parameters", 400 unless [:username, :email, :password].none? { |x| params[x].blank? }
-    return error! "User with that email already exists", 409 if User.exists?(['lower(email) = ?', params[:email].downcase])
-    return error! "User with that name already exists", 409 if User.exists?(['lower(name) = ?', params[:username].downcase])
+    return error! 'Missing parameters', 400 unless %i(username email password)
+                                                   .none? do |x|
+                                                     params[x].blank?
+                                                   end
+    return error! 'User with that email already exists', 409 if User.exists?(
```
NuckChorris: These should probably get offloaded to model methods, but if you don't wanna bother right now this works well enough

app/controllers/auth_controller.rb
Eschon: How would I do that? Won't rubocop complain when I change it to
```Ruby
unless [:email, :password].none? { |x| params[x].blank? }
  return error! 'Missing parameters', 400
end
```

NuckChorris: I _hope_ it's smart enough to tell that it would violate another rule to flip it… but I don't know

Eschon:
I changed the long conditionals like @NuckChorris suggested and rubocop was ok with it.
I squashed everything so feel free to merge if nothing else needs to be changed.

NuckChorris:
:+1: You're really tearing through these! LGTM!

Eschon:
I don't have that much free time but sometimes I still don't know what to do with it :sweat_smile:

PR #606, opened 2015-07-20
[WIP] Add Mangas status to feed

DNA:
This is a WIP PR to add manga updates in the user feed. By now it add every reading status change to the timeline. The next step will be to add reading chapters. And some tests :)

### DIFF
--- a/app/services/action.rb
+++ b/app/services/action.rb
@@ -65,6 +65,20 @@ class Action
     end
   end
 
+  def self.manga_status(m)
+    return unless m.valid?
+
+    Substory.create({
+      user: m.user,
+      substory_type: Substory.substory_types[:manga_status_update],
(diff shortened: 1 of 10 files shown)


JoshFabian:
You're my g'damn hero.

DNA:
How manga are being displayed by now: 
screen shot 2015-07-20 at 1 35 11 am

app/services/action.rb
```diff
@@ -65,6 +65,20 @@ def self.broadcast(data)
     end
   end
 
+  def self.manga_status(m)
```
NuckChorris: This goes completely against the current setup and I completely approve of it.  The current `#broadcast` method is a disaster, and this avoids it.  :+1:

PR #607, opened 2015-07-20
Remove the Watchlists model

NuckChorris:
There's been a lot of legacy crap using the Watchlists model, which for some damn reason never got properly torn out.  This tears it out.

Oh and it also adds a way for MessageFormatter to extract mentioned users, which I'll be implementing as part of my feed refactoring ;)


app/models/library_entry.rb
```diff
@@ -33,6 +33,14 @@ class LibraryEntry < ActiveRecord::Base
   VALID_STATUSES = ["Currently Watching", "Plan to Watch", "Completed", "On Hold", "Dropped"]
   validates :status, inclusion: {in: VALID_STATUSES}
 
+  SNAKE_PARAMETERS = {
```
DNA: I don't like this constant name because it don't relate to the hash content. maybe something like `SNAKE_STATUSES` or `STATUSES_HASH`?

app/models/story.rb
```diff
@@ -41,14 +41,20 @@ class Story < ActiveRecord::Base
 
   def self.for_user_and_anime(user, anime, story_type="media_story")
     story = user.stories.where(story_type: story_type, target_id: anime.id, target_type: "Anime")
-    watchlist = Watchlist.find_by(user_id: user.id, anime_id: anime.id)
+    entry = LibraryEntry.find_by(user_id: user.id, anime_id: anime.id)
```
DNA: To remove that if, what you think about what I did on https://github.com/DNA/hummingbird/blob/add-mangas-to-feed/app/models/story.rb#L57 ??

NuckChorris: Yeah that's on the list for my `feed-refactor` branch for this week

app/models/library_entry.rb
NuckChorris: I can agree with that, I kind of halfassed the naming.  Will fix in morning.

app/models/story.rb
```diff
@@ -70,11 +76,10 @@ def set_last_update_time!(time=nil)
   end
 
   def self.for_user(user)
-    if user.nil? or user.sfw_filter
-      where("NOT adult").joins("LEFT OUTER JOIN watchlists ON watchlists.id = stories.watchlist_id").where("watchlists.id IS NULL OR watchlists.private = 'f'")
-    else
-      joins("LEFT OUTER JOIN watchlists ON watchlists.id = stories.watchlist_id").where("watchlists.id IS NULL OR watchlists.private = 'f'")
-    end
+    query = eager_load(:library_entry)
+            .where("library_entries.id IS NULL OR library_entries.private = 'f'")
```
DNA: I think we can go further removing the sql with 
```ruby
.where(library_entries: { id: nil })
.where_or(library_entries: { private = 'f' })
```
what do you think?

app/services/message_formatter.rb
```diff
@@ -2,6 +2,7 @@
 require 'onebox'
 
 class MessageFormatter
+  MENTION_REGEXP ||= /@[_A-Za-z0-9]+/
```
DNA: I sure to like this! Easier to understand :+1:

app/services/message_formatter.rb
```diff
@@ -24,6 +25,16 @@ def self.format_message(message)
     MessageFormatter.new(message).format
   end
 
+  def self.extract_mentions(message)
+    MessageFormatter.new(message).mentions
```
DNA: If we initialize the method each time we call an static method, why won't use a singleton instead?

test/controllers/library_entries_controller_test.rb
```diff
@@ -79,7 +79,7 @@ class LibraryEntriesControllerTest < ActionController::TestCase
   end
 
   test "need to be authenticated as correct user to update library entry" do
-    id = watchlists(:one).id
+    id = library_entries(:one).id
```
DNA: since we are refactoring this, can't we change this to a factory?

DNA:
Nice refactor! It's much more understandable now :)

app/services/message_formatter.rb
NuckChorris: Because ideally this shouldn't be static methods.  I just did this to be consistent for now.  Eventually the MessageFormatter instance should be the direct interface, not this

test/controllers/library_entries_controller_test.rb
NuckChorris: I have better things to do with my time than refactor already-working tests :wink:

app/models/story.rb
NuckChorris: My understanding is that `or` is not possible in the ActiveRecord::Relation DSL, that you have to drop down to ARel or SQL for it.  Also it would be `private: false` since `'f'` is just the way they do false in Postgres

DNA: damn, I was pretty sure that there was a method to do this already… Unfortunately, we'll need to wait til Rails 5 for it :/

NuckChorris: > Unfortunately, we'll need to wait til Rails 5 for it

IDK if it's even in Rails 5.  I'm thinking of pulling in Squeel.  It'd make some of our common queries (`LIKE`, `>=`, etc.) so much cleaner

DNA: I'm not sure if I like this sintax… :P 
But there'll be an `or` method on Rails 5, I've checked already: https://github.com/rails/rails/pull/16052

NuckChorris: I can't say I'm terribly fond of the syntax either, but it beats the hell out of SQL

vevix:
:100: LGTM

PR #609, opened 2015-07-25
resolve rubocop offenses for characters_controler.rb

Eschon:
Hey guys I did a big one again :stuck_out_tongue_winking_eye: 
btw do you check new PRs before merging them? 
Would be a shame if all the work fixing this would go to waste.


NuckChorris:
Did you make sure to write an extensive unit test suite to ensure that this change won't break anything? We obviously can't accept it without 😜

LGTM!

PR #611, opened 2015-07-27
Refactor minitest

DNA:
That messy output that was generated when running tests was getting on my nerve, so I refactored it. I also added support to Rconv, so we can check code coverage data that is generated on the `/coverage` folder

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 require File.expand_path('../boot', __FILE__)
 
 require 'rails/all'
+require 'coffee_script'
 
 # require_dependency doesn't work here for some reason.
 require_relative '../lib/log_before_timeout.rb'
--- a/lib/tasks/test_directories.rake
+++ b/lib/tasks/test_directories.rake
@@ -1,16 +1,13 @@
 namespace :test do
-  task run: %w(
-    test:units test:functionals test:generators test:integration
-    test:services test:libs
-  )
-
   desc 'test services'
   Rails::TestTask.new(services: 'test:prepare') do |t|
     t.pattern = 'test/services/**/*_test.rb'
   end
 
-  desc 'test libs'
-  Rails::TestTask.new(libs: 'test:prepare') do |t|
+  desc 'test lib'
+  Rails::TestTask.new(lib: 'test:prepare') do |t|
     t.pattern = 'test/lib/**/*_test.rb'
   end
 end
+
+Rake::Task['test:run'].enhance ['test:lib', 'test:services']
--- a/lib/tasks/test_directories.rake
+++ b/lib/tasks/test_directories.rake
@@ -1,20 +1,16 @@
 namespace :test do
-  desc "test lib"
-  Rake::TestTask.new(:lib) do |t|
-    t.libs << "test"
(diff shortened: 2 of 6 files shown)


config/application.rb
```diff
@@ -1,6 +1,7 @@
 require File.expand_path('../boot', __FILE__)
 
 require 'rails/all'
+require 'coffee_script'
```
NuckChorris: We should probably actually fix this so we don't have coffeescript in our HAML

lib/tasks/test_directories.rake
```diff
@@ -1,20 +1,16 @@
...
-  desc "test lib"
-  Rake::TestTask.new(:lib) do |t|
-    t.libs << "test"
-    t.pattern = 'test/lib/**/*_test.rb'
-    t.verbose = true
-  end
+  task run: %w(
+    test:units test:functionals test:generators test:integration
+    test:services test:libs
+  )
```
NuckChorris: I made these changes already in https://github.com/hummingbird-me/hummingbird/commit/a800e1c48af79d49b95ed277fd0de67a47e7ca6f already on the oauth branch

test/test_helper.rb
```diff
@@ -1,45 +1,78 @@
...
+    add_filter '/vendor/'
+
+    add_group 'Controllers', 'app/controllers'
+    add_group 'Models', 'app/models'
+    add_group 'Services', 'app/services'
+    add_group 'Serializers', 'app/serializers'
+    add_group 'Libs', 'lib/'
+  end
+  SimpleCov.formatter = SimpleCov::Formatter::HTMLFormatter
+end
```
NuckChorris: The coverage stuff is getting a bit hefty with these changes, would it be possible to break it into a separate file?

test/test_helper.rb
```diff
@@ -1,45 +1,78 @@
...
+require 'webmock/minitest'
+WebMock.disable_net_connect!(allow: %w(codeclimate.com robohash.org))
 $redis = ConnectionPool.new { MockRedis.new }
 
-class ActiveSupport::TestCase
-  fixtures :all
-  include FactoryGirl::Syntax::Methods
-end
+module ActiveSupport
+  class TestCase
```
NuckChorris: I was gonna ask why you were nesting with a module but after some research I now realize we should be using this style more.  There's some strange constant-scoping stuff involved in `Foo::Bar` namespacing

>In the case of class Foo::Bar it really should push TWO scopes onto the stack (both Foo and Bar), but it only pushes one, hence we get the "problem"

config/application.rb
DNA: I thought so, but it goes beyond the scope of this PR. We can create an issue to fix this

lib/tasks/test_directories.rake
DNA: Ok, used the same code, so we can prevent any conflict :)

test/test_helper.rb
DNA: Good to know. I was only following rubocop, didn't know it has it scope problem

NuckChorris:
LGTM!

PR #612, opened 2015-07-30
allow unrating on media pages. fixes #595

### LINKED ISSUE #595: Bug: Unable to un-rate on anime/manga pages.
You can unrate a show by giving the anime the same rating on the library page (`https://hummingbird.me/users/wopian/library`)

However, you are unable to unrate a show on the anime page (`https://hummingbird.me/anime/overlord`)

Same for manga.

@NuckChorris

vevix:
:speech_balloon: :arrow_up:

### DIFF
--- a/frontend/app/components/awesome-rating.js
+++ b/frontend/app/components/awesome-rating.js
@@ -8,14 +8,15 @@ export default Ember.Component.extend({
   media: null,
 
   applyAwesomeRating: function() {
-    var self = this;
     this.$().AwesomeRating({
       rating: this.get('rating'),
       type: this.get('type'),
       editable: this.get('editable'),
-      update: function(newRating) {
-        self.sendAction('action', newRating, self.get('media'));
-        self.set('rating', newRating);
+      update: (newRating) => {
+        if (this.get('rating') === newRating) {
+          newRating = null;
+        }
+        this.sendAction('action', newRating, this.get('media'));
       }
     });
   }.on('didInsertElement').observes('rating')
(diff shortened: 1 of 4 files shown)


NuckChorris:
Should this be in the rating widget instead?

vevix:
I think it might be better to create some media mixins for this sort of functionality between manga/anime instead of all the duplicate code.

This fits with how it is handled on the library page currently.

NuckChorris:
I was thinking more like making the Awesome Rating widget return null if it's the same value

vevix:
We would need to change it up a fair bit as it's used on the onboarding pages before library entries are created.

NuckChorris:
Can't we compare the clicked value with the displayed value, and not have it refer at all to the library entry?

vevix:
:100: That could work!

vevix:
:bulb: bump.

Looking forward to seeing how the library handles on 1.13 & glimmer instead of react.

frontend/app/components/awesome-rating.js
```diff
@@ -8,14 +8,15 @@ export default Ember.Component.extend({
...
       type: this.get('type'),
       editable: this.get('editable'),
-      update: function(newRating) {
-        self.sendAction('action', newRating, self.get('media'));
-        self.set('rating', newRating);
+      update: (newRating) => {
+        if (this.get('rating') === newRating) {
+          newRating = null;
+        }
+        this.sendAction('action', newRating, this.get('media'));
```
NuckChorris: Will the loss of `self.set('rating', newRating)` change behavior at all?

vevix: Not with the way we currently use it. `rating` is bound to the library entry so when it gets updated via the sent action, it reflects here.

NuckChorris:
LGTM! 🚀

PR #613, opened 2015-07-30
fix image link for pro-banner on users page

vevix:
:speech_balloon: :arrow_up:

Relative made it so on pages such as `/library` broke the banner link.


NuckChorris:
Ive looked at this error like 70times, but always forgot to fix it

:+1: LGTM

PR #615, opened 2015-08-11
Enhance quickupdate when viewing on half of a 1080p screen

FlashYoshi:
Some css changes that allow people viewing the site on half of a 1080p screen to see and use the arrows (most handy!) on the quick update view. It also disables the scrollbars when they are not needed instead of greying them out.


JoshFabian:
Good stuff, thanks for this!

FlashYoshi:
Awesome reaction speed @JoshFabian!

DNA:
That was the quickest PR I ever saw here on HB :O

Issue #617, opened 2015-08-12
APIv1 Feeds give 500 error


NuckChorris:
Seems to be a polymorphic eager loading error, maybe regression of #539 (solved in 0f3740e9), presumably triggered by #607

```
ActiveRecord::EagerLoadPolymorphicError (Cannot eagerly load the polymorphic association :target):
 app/api/api_v1.rb:318:in `block (2 levels) in <class:API_v1>'
```

DNA:
PR fix sent, see if it's okay

PR #618, opened 2015-08-12
Fix eager loading not using left outer join

DNA:
`eager_loading` and `joins` can't create a `LEFT OUTER JOIN`, so it brings more data than we need, getting some problems like #617. This PR fixes it.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -76,7 +76,7 @@ class Story < ActiveRecord::Base
   end
 
   def self.for_user(user)
-    query = eager_load(:library_entry)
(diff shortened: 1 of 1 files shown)


DNA:
I've sent it without tests as a hotfix, but I think we need to think about some tests on feed to prevent this in the future, what you guys think?

NuckChorris:
If you look at my feed-refactor branch you'll see that I'm adding unit tests in the process

NuckChorris:
:pray: :100: :fire: :boom: :droplet: LGTM

PR #625, opened 2015-08-13
Try removing XMLCleaner

NuckChorris:
Replace with `String#scrub`

This might backfire, but currently there's an `xmllint` issue with `EPIPE` where xmllint appears to be closing its pipe early.  It seems like it's worth a try, now that we have at least _some_ error handling.

### DIFF
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -20,7 +20,7 @@ class Settings::ImportController < ApplicationController
     else
       return error!(400, 'Unknown format')
     end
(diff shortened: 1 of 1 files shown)


DNA:
I'm always in for changing some personal code into native code, let's test it! :+1:

DNA:
And if some xml breaks, we can put it in our tests in the future :P

NuckChorris:
This actually does less than `XMLCleaner` did, but I'm not certain that they're handing us invalid XML, just that they're handing us invalid encodings.

vevix:
:+1: Also :100: branch :smirk:

PR #626, opened 2015-08-13
Fix synopsis remover in importer

synthtech:
MAL updated their "No synopsis" message again.

Ex.
https://hummingbird.me/manga/koi-wa-ameagari-no-you-ni
https://hummingbird.me/manga/youkoso-jitsuryoku-shijou-shugi-no-kyoushitsu-e


NuckChorris:
Why does xinil have to make it so hard to steal his data!? :+1: LGTM

PR #628, opened 2015-08-14
Update Headers

pixeldesu:
_Fixes issue mentioned in #624_ and also fixes/improves some other things I noticed.

To sum up the changes:

* Removed "Profile" links from the Dashboard dropdown, as it becomes unnecessary when there already is "View Profile" in the profile dropdown. (both headers)
* Renamed Dashboard to Library as it is only holding links to the libraries and recommendations and the Dashboard is generally accessible over the navbar brand. (both headers)
* Removed the link from the Dashboard dropdown in the old header design to unify it with the new header, as dropdown links in Bootstrap can't open another page while being clicked. (basically just links to `#`)
* Added Manga library link (old header)
* Added "View Profile" and renamed "Edit Profile" to "Settings" in profile dropdown (old header)

As the old header is going to be dropped at some point these changes are just temporary solutions until everything is switched over to the new design and implemented.

### DIFF
--- a/app/views/shared/_header.html.haml
+++ b/app/views/shared/_header.html.haml
@@ -12,12 +12,12 @@
           / Left Nav Section
           %ul.left.topnav-links
             %li.has-dropdown
-              = link_to "Dashboard", user_signed_in? ? "/" : "/sign-in"
+              = link_to "Library", "#"
               %ul.dropdown.rounded-panel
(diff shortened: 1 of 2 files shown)


JoshFabian:
LGTM! Definitely agree with the dashboard / library semantics.

PR #629, opened 2015-08-15
fix rubocop offenses in episodes_controller.rb

Eschon:
I fixed some offenses again.
I hope it's not dead code again :sweat_smile:


vevix:
This one is only minor changes but great work! This code is the opposite, where we are planning to use it in the future.

PR #630, opened 2015-08-15
fix quick-update pagination being enabled with 0 entries on next page

pixeldesu:
_this is pretty much just an attempted fix (as I can't test locally right now) for one of the issues mentioned in #152_

if the line is too long I can move parts of it into a new function, but I'd like to know if the code itself works first, so I need someone to take a quick glimpse at it.

### DIFF
--- a/frontend/app/components/quick-update.js
+++ b/frontend/app/components/quick-update.js
@@ -7,7 +7,7 @@ export default Ember.Component.extend({
   libraryEntries: [],
   canGoBack: Ember.computed.gt('page', 1),
   canGoForward: function() {
-    return !this.get('loading') && this.get('libraryEntries.length') === 6;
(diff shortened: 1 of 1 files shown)


vevix:
This would fix the issue, but I dont like it because we would now be issuing another network request without cleaning up the code to actually use the fetched data. 

It might be a better approach to return 7 entries so we can peak into the next row.

pixeldesu:
We could go for something like a `maxpage` value, which is simply put together with the amount of library entries divided through 6 and then rounded up the next number if it's a decimal.

Then we just need to know the current page position which needs to be lower or exactly the `maxpage` value.

vevix:
:+1: This approach would be better.

DNA:
:+1: for `maxpage` value :P

PR #633, opened 2015-08-24
Add some damned margins to the footer links

NuckChorris:
screen shot 2015-08-23 at 7 41 59 pm

This has bothered me long enough


vevix:
:+1: I like this change!

Issue #623, opened 2015-08-13
APIv1 anime objects have empty title values


fractalemagic:
_Some_ anime objects returned by APIv1 have an empty `title` value when English is selected as the primary language.

Excerpt from [api/v1/anime/durarara?title_language_preference=english](https://hummingbird.me/api/v1/anime/durarara?title_language_preference=english)

```
{
  "id": 4696,
  "slug": "durarara",
  "title": "",
  "alternate_title": "Durarara!!",
```

Other affected titles include [aldnoah-zero-2](https://hummingbird.me/api/v1/anime/aldnoah-zero-2?title_language_preference=english), [hunter-x-hunter-2011](https://hummingbird.me/api/v1/anime/hunter-x-hunter-2011?title_language_preference=english), [psycho-pass](https://hummingbird.me/api/v1/anime/psycho-pass?title_language_preference=english), [eureka-seven](https://hummingbird.me/api/v1/anime/eureka-seven?title_language_preference=english)

erengy:
Can be reproduced by setting the `title_language_preference` parameter to `english`:
https://hummingbird.me/api/v1/anime/4696?title_language_preference=english

Probably related to 0a3c19a10bb320e73ee8b9658748cadfa75e2a68, or more specifically, [this line](https://github.com/hummingbird-me/hummingbird/blob/31e10981feb9c568552dc4672c157674cbe65c8b/app/models/anime.rb#L162). I don't know Ruby, but is this because `alt_title` is empty string rather than `nil` for these series?

fractalemagic:
Ah. I didn't know that you could supply the preference language like that, thanks.

DNA:
I've looked at this issue and @erengy was right on the exactly problematic line. It happens because there's some titles there is just an empty string on the database, so rails think it's a valid string. Unfortunately this fact slipped though me when I refactored the code, but I'll send a fix for it asap.

z411:
Indeed, it appears those are shows that have an empty title in the database, because in the official web site they also appear as empty when searching for them. If that's the case, then the API is not really broken in this regard and should be fixed in the database instead?

![be8646bc-53e1-11e5-9fcb-153471bc4ea0](https://cloud.githubusercontent.com/assets/1988512/9778724/b8684ba6-574a-11e5-98b5-574ab6c9d010.png)

erengy:
I think there's still an issue with `canonical` title preference.

URL|title
---|:---:
[...]

PR #636, opened 2015-09-11
fix titles for english preference

vevix:
REF: #623 

Anything else needed? We should get this pushed ASAP /cc @DNA

### DIFF
--- a/app/models/anime.rb
+++ b/app/models/anime.rb
@@ -159,7 +159,7 @@ class Anime < ActiveRecord::Base
     when 'canonical'
       english_canonical? ? alt_title : title
     when 'english'
-      alt_title || title
+      alt_title.present? ? alt_title : title
     else
       title
     end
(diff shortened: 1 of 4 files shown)


vevix:
Also fixed what seems to be a bug reported here: https://forums.hummingbird.me/t/canonical-titles-only-showing-as-romanized/26846

Our title code is pretty bad, I think @NuckChorris was working on a change for it :(.

NuckChorris:
I was hoping to refactor it into a separate table but thats more of a long term endeavor and between my feed refactor and oauth2 I've already got enough patches to stage.

For now I think renaming them english_title and romaji_title would help greatly.

DNA:
This should resolve all the present bugs until do a better refactor :)

DNA:
Maybe add some tests to check if we are not presenting empty string as valid ones? It'll help in the next refactor too

Issue #634, opened 2015-08-24
Entering anime rating 0.0 through the web interface causes 'Error connecting to the server'.


nstratos:
When using the advanced rating system and when trying to enter 0.0 stars as rating (it's a little tricky with the mouse), a message pops up saying: 'Error connecting to the server' and keeps retrying to submit the value indefinitely. The real reason is that the server responds with: 
```json
"{"error":"Rating is not in the valid range"}"
```

When entering 0.0 rating through the API, the value is submitted without error. I assumed that the web interface would also be able to accept 0.0 rating but this does not seem to be the case.

Is this the intended behavior?

nstratos:
It seems that in if you give an anime the same rating as it originally had, it actually reduces the rating to zero. I was not aware of that.

It might still be reasonable to change the error message (Error connecting to the server) to something more descriptive so that less users get confused.

vevix:
>It seems that in if you give an anime the same rating as it originally had, it actually reduces the rating to zero. I was not aware of that.

This is by design, as this is how we let users unrate a anime, or manga.

--------------

I didn't even know it was possible to select 0.0, that seems weird.and something we should fix.

nstratos:
It's a little tricky to select 0.0 through the web interface by using the mouse so it might not be that big of a problem. In any case, I created a little gif to help.

![zero-rating-hummingbird](https://cloud.githubusercontent.com/assets/3655711/9542342/8f3a1dd0-4d78-11e5-9b29-024935e8d44e.gif)

vevix:
Thanks for the GIF :D that's crazy.

vevix:
Thanks this was fixed in #637 :kissing:

PR #637, opened 2015-09-11
don't allow 0 value ratings

vevix:
REF: #634

This matches the functionality of APIv1. (Where 0 also == reset)


NuckChorris:
:+1: it's a rare edge case but it should be covered

PR #638, opened 2015-09-11
Update Manga Ratings

vevix:
Should be good based off internal testing. Did you need to do something about previous ratings @NuckChorris?, I remember a forum post about something like that.


NuckChorris:
Yeah I still need to fill the manga rating buckets with existing ratings. Pain in the ass it is

PR #639, opened 2015-09-11
update onebox & add spotify engine

vevix:
This has been a pretty well [requested](https://forums.hummingbird.me/t/spotify-embedding/25276) feature  & one we have talked about in Slack before. Also :arrow_up:'s Twemoji due to Nokogiri dep.

screen shot 2015-09-11 at 10 19 22 pm

-----

Once we get the admin status worked out for the repo it might be a good idea to fork onebox here and push some of these upstream eventually.


darth62969:
why... like why...

PR #643, opened 2015-09-15
Fix importing of genres and ratings for anime

synthtech:
It looks like there was a slight change in the structure of anime pages on MAL which broke some things.
Ex. https://hummingbird.me/anime/pokemon-xy-z

### DIFF
--- a/lib/mal_import.rb
+++ b/lib/mal_import.rb
@@ -111,7 +111,6 @@ class MALImport
       rescue
       end,
       poster_image: poster_image(@sidebar.css("img")[0]['src']),
-      genres: begin (@sidebar.css('span:contains("Genres:") ~ a').map(&:text) rescue []).compact end,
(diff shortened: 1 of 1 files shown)


vevix:
After testing I'm still seeing some issues with the age rating.

```ruby
:age_rating=>"Rating:\n  R - 17+ (violence & profanity)",
:age_rating=>"Rating:\n  PG - Children",
```

This should be fixed by changing the text search to `"Rating:\n "` instead of `"Rating: "`

synthtech:
@vevix Thanks, fixed it

I also just noticed now that the episode count and duration might be broken as well. Should I just add another commit here or make a new PR?

vevix:
Here is fine :+1:

synthtech:
One last thing, the text describing the duration for TV series was changed to "min. per ep." rather than just "min." and I don't know if that would break the regex under`parse_duration`.

vevix:
Thanks :+1: I need to make some edits and I'll manually merge this.

synthtech:
Ok, great :smile:

PR #646, opened 2015-09-21
MAL Import fixes

vevix:
Okay so there was a few issues.

First: We were calling `raise` on a captured `Exception` and therefore the error text was never sent to the client so their status was never changed from queued and got stuck there.

Second: Seems the file was in `binmode` which uses ASCII-8BIT encoding so Sidekiq cried when passing it as an argument.

--------------

We're going to need to do something about people who are stuck in queued status, this includes users who were stuck because of the above bug, and users like @wopian who were bugged from the migration update.

### DIFF
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -34,9 +34,11 @@ class Settings::ImportController < ApplicationController
 
     mixpanel.track 'Imported from MyAnimeList', {email: current_user.email} if Rails.env.production?
   rescue Exception
-    error! 500, 'There was a problem importing your list.  Please send an email
-                 to josh@hummingbird.me with the file you are trying to import
-                 and we\'ll see what we can do.'
-     raise
+    status = User.import_statuses[:error]
+    current_user.update_columns import_status: status,
+      import_error: 'There was a problem importing your list. Please send an email
+                    to josh@hummingbird.me with the file you are trying to import
+                    and we\'ll see what we can do.'
+    raise
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/settings/import_controller.rb
```diff
@@ -34,9 +34,11 @@ def myanimelist
...
-    error! 500, 'There was a problem importing your list.  Please send an email
-                 to josh@hummingbird.me with the file you are trying to import
-                 and we\'ll see what we can do.'
-     raise
+    status = User.import_statuses[:error]
+    current_user.update_columns import_status: status,
+      import_error: 'There was a problem importing your list. Please send an email
+                    to josh@hummingbird.me with the file you are trying to import
+                    and we\'ll see what we can do.'
+    raise
```
NuckChorris: So basically if something went wrong before the Sidekiq task was created, it would never reset them to an unerrored state?

Damn, I looked at this code like 30 times trying to find the hole and never noticed that

We should probably still `error!` to pitch it down to the client immediately though, right?

frontend/app/models/current-user.js
```diff
@@ -32,28 +32,16 @@ export default User.extend({
...
-          importStatus: 'error',
-          importError: err.jqXHR.responseJSON.error
-        });
-      } catch (e) {
-        alert("There was a problem importing your list. Send an email to josh@hummingbird.me with the file you're trying to import and we'll see what we can do.");
-      }
+    }).then((data) => {
+      this.store.pushPayload(data);
+    }, () => {
+      window.location.reload();
```
NuckChorris: Hah, brilliant.  It's ghetto but it works well enough

NuckChorris:
I can unstuck them with a quick query on the server, also fuck Sidekiq.

Anyways :+1: looks good

And makes me feel dumb for not catching the edgecase there

PR #682, opened 2015-10-11
Docker support for the future version

### LINKED ISSUE #654: Setup Docker
Create Dockerfile's for development purposes.

MiLk:
Fixes #654 

# SSH Key

You need to add a ssh key which can access this repo: https://github.com/hummingbird-me/foundation-sites-6
Reuse an existing key, or create a new key that you will use only for github, and save it in `github_key` in the project's root directory
It will be copied inside your docker container. 

# General instructions

```bash
docker-compose build
docker-compose up web
```

## Linux

```bash
xdg-open "http://localhost:3000"
xdg-open "http://localhost:$(docker-compose port web 4200 | cut -d':' -f2)"
```

## OSX

```bash
open "http://$(docker-machine ip docker-machine-dev):3000"
open "http://$(docker-machine ip docker-machine-dev):$(docker-compose port web 4200 | cut -d':' -f2)"

```


vevix:
Ran into an issue with the installation of `node-sass`. 

```bash
Error: The `libsass` binding was not found in /opt/hummingbird/client/node_modules/ember-cli-sass/node_modules/broccoli-sass-source-maps/node_modules/node-sass/vendor/linux-x64-14/binding.node
This usually happens because your node version has changed.
Run `npm rebuild node-sass` to build the binding for your current node version.
```

Phantomjs 2.0 should also be installed for running the emberjs tests.

MiLk:
Did you install the dependencies on your system before?
Try to remove all the `node_modules` directory and try again.

vevix:
Yep, that & not using VirtualBox 5.0.4 (due to a symlink bug) worked.

MiLk:
More work required on this one?
poke @JoshFabian @NuckChorris

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    POSTGRES_PASSWORD: mysecretpassword
+    POSTGRES_USER: postgres
+redis:
+  image: redis
+web:
+  build: .
+  volumes:
+    - ./:/opt/hummingbird/
+  ports:
+    - 3000
```
vevix: Can this be changed to `3000:3000`? 

Need to know the port the API is running on to send AJAX requests from the browser.

MiLk: It was to avoid conflicts with other applications.
I changed that.

Do you want 4200 to be fixed too?

vevix: Nah that one doesn't matter.

vevix:
:+1: from me, Will wait for @NuckChorris to sign off also.

NuckChorris:
What about the build? It can't find JJ Abrams

vevix:
Will need to rebase against master to fix that specific fail, ESA released
1.0 and deleted that branch.

PRs are failing ATM due to Travis not exposing secure environment variables
(private key) to PRs from forks.

MiLk:
You mean against `the-future`, right?

vevix:
Oops, yeah `the-future`

NuckChorris:
Shit can we expose the key anyways? It's just to the one repo and it's not like it's a big security thing. Worst case scenario, somebody sees F6 a few months early

vevix:
We can set the key to public which will allow PRs to use but it will also log the key in every build log. If we're okay with that :/ I can make the change.

NuckChorris:
We can't make it secure-but-present-in-PRs? I guess it is kind of a really niche thing

vevix:
Nope, doesn't seem possible. 

>Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

I looked into storing the key in the repo encrypted, but travis stores the encryption keys as secure environment variables so that doesn't work either.

NuckChorris:
Is the key read-only? If so I'd say make it public. Otherwise we could be 'sploited easy

vevix:
Yep, it's read-only. and now public.

NuckChorris:
Well anyways, back on topic: this is a +1 from me, the solution here seems pretty viable. Still seems silly to me that we have to do this whole rigamarole but w/e

NuckChorris:
Eventually we're gonna wanna split them into separate Dockerfiles, mind you. But this is a start, and we desperately need a simple dev environment like this

MiLk:
I agree with splitting in multiple dockerfiles, but it would mean a different foreman config or no foreman at all.

NuckChorris:
We don't necessarily need to use Foreman

NuckChorris:
Technically speaking I think when you deploy a Procfile-based app on Heroku it splits each into a separate LXC Container

MiLk:
Updated. Test passed.

NuckChorris:
Looks good to me

vevix:
Would be good to get the wiki information updated if needed.

MiLk:
Will do it when it becomes the main branch.
Makes sense?

vevix:
:+1: I'll get this merged as soon as I can finish up some Ember code needed to support docker. Hopefully tonight.

JoshFabian:
Is this now the preferred way for a new local machine to be setup? Can we update the readme to reflect this? Lot of new contributors interested in helping that this will be helpful for.

MiLk:
Hum, which place? The readme is already pointing to the right page of the wiki.

Issue #673, opened 2015-10-04
Badge System


JoshFabian:
Allow users to add badges while using the site. We can begin listing potential badge ideas within this issue, deciding on an initial set of badges to start with when development begins.

This could potentially be done with this rails gem: https://github.com/merit-gem/merit

NuckChorris:
> Copied from #684 and #686

# Achievement Badges

Hummingbird would like to be able to reward active users for using the site in a way that allows them to
show off their reputation in the community as well as the variety of anime and dramas they have watched,
manga they have read, etc.  Hummingbird's badges could be similar in style to 
[those of MALGraph](https://github.com/rr-/malgraph4/wiki/Achievements).

In this document, we describe the categories of badges available, as well as the specific badges in each
category, how they are unlocked, and what they represent.  This guide is meant to serve largely as a
developer-first specification of how badges will be implemented, but in such a way that user documentation
(in the form, for example, of a blog post) could be easily written.

## Categories

### Anime

Title | Description | How to Unlock
------|-------------|--------------
First Franchise | Completed an entire franchise | Watched every episode of anime in a franchise
Anime Stalker | Start to expect anime shows | Plan to watch 5 shows and start watching them after they begin airing.
New comer | Welcome to the anime world | Total time of watching anime accumulated to 3 days
Engaged | Feel a sense of attachment | Total time of watching anime accumulated to 7 days
Anime Pursuit | Anime goes beyond daily life | Total time of watching anime accumulated to 1 month
Otaku Soul | Anime merged to your Otaku soul | Total time of watching anime accumulated to 3 month
Challange Accepted | You shall never leave the world | Total time watching anime accumulated to 1 year
Feel elated | Comedy is really your fav | Watched 50 titles of comedies
Love life | Romantic anime are the besto | Watched 50 titles of romance
Schooler | School life memory | Watched 50 titles of school
Nonnon Biyori | Enjoy the life | Watched 50 titles of slice of life
Fighting | With your samurai spirit | Watched 50 titles of action
Drama Goer | Drama is life | Watched 50 titles of drama
El Psy Congroo | We love science | Watched 50 titles of Sci-Fi

### Manga

Title | Description | How to Unlock
------|-------------|--------------
Heavy Reader | Completed 25 manga | Manga count reaches 25


### Drama

Title | Description | How to Unlock
------|-------------|--------------
Dramaniac | Watched ALL the dramas | Drama count = number of dramas in DB

### Community

[...]

PR #684, opened 2015-10-11
Specification document for badge system

zsck:
See the [relevant issue](https://github.com/hummingbird-me/hummingbird/issues/673)

[Rendered](https://github.com/redwire/hummingbird/blob/badges-spec/docs/badges.md)

**NOTE** This PR does not need to be merged right away.  I am creating it so that we have a place to discuss and increment on the badge system specification.


vevix:
Great start! :100:. If we're going to store these specs in this repository I would prefer them to be under `docs/` over `doc/`.

cybrox:
Since we have a base to collect all the notes with this, it might be worth going over all the suggestions in the [Old Forum Thread](http://forums.hummingbird.me/t/what-achievements-should-hummingbird-have/6332) again and pick the useful ones out.

NuckChorris:
Would it be better to collect RFCs in issues with a label? That way we can see the markup'd version

redwire:
I think it's a great idea to have an issue for each RFC type of thing.
Having a pull request lying around gives us a lot more information, though. E.g. you can see commits (whose lines can be commented on), and so on.

docs/badges.md
```diff
@@ -0,0 +1,38 @@
...
+
+Hummingbird would like to be able to reward active users for using the site in a way that allows them to
+show off their reputation in the community as well as the variety of anime and dramas they have watched,
+manga they have read, etc.  Hummingbird's badges could be similar in style to 
+[those of MALGraph](https://github.com/rr-/malgraph4/wiki/Achievements).
+
+In this document, we describe the categories of badges available, as well as the specific badges in each
+category, how they are unlocked, and what they represent.  This guide is meant to serve largely as a
+developer-first specification of how badges will be implemented, but in such a way that user documentation
+(in the form e.g. of a blog post) could be easily written.
```
wopian: (e.g. in the form of a blog post) reads better. "E.g" sounds awkward in the middle of the sentence.

docs/badges.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Categories
+
+### Anime
+
+Title | Description | How to Unlock
+------|-------------|--------------
+First Franchise | Completed an entire franchise | Watched every episode of anime in a franchise
```
wopian: "of **an** anime"

NuckChorris:
That's true, the ability to comment on lines is pretty damn useful. This'll work.

So I agree with the concept so far, seems good. Anyone wanna go through the list of MALgraph badges and see which ones we should steal?

docs/badges.md
redwire: @wopian This one's not referring to just a single anime in a franchise, but all the anime in that franchise.

andyx96:
I can go through all the anime genres from MALgraph, maybe modify them a little bit to cater the hummingbird categorization.

vevix:
Updated the OP to include a rendered file link.

docs/badges.md
wopian: :+1:

NuckChorris:
I'm just merging this into #673 because we don't need two open PRs and an issue just to discuss the spec.  A single issue works fine too.

PR #686, opened 2015-10-14
Badges

zsck:
This is another placeholder PR. It will be developed behind my `badge-spec` branch. Thus, once all the badges are specified, `badge-spec` should be merged **first**.  Then this branch will grow to merge changes from `the-future` and anything in `badge-spec` as well as contain badge code.


NuckChorris:
I'm just merging this into #673 because we don't need two open PRs and an issue just to discuss the spec. A single issue works fine too.

PR #690, opened 2015-11-01
This commits pretend to clear a little bit the user model

nivla:
I don't know if you agree guys, but I think we should clear some models and follow some
rules and styles from the guideline,  I see some fat models and large methods that it hard to see, understand and mantain.

We could decouple some large methods into pieces and add some test for them.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -68,56 +68,13 @@
 #
 
 class User < ActiveRecord::Base
-  # Friendly ID.
-  def to_param
-    name
-  end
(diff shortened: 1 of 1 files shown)


NuckChorris:
We're no longer doing or accepting cleanup on `master`, due to the possibility of breakage.  `master` is now effectively frozen — only bugfixes will be merged.

We're working on `the-future` now and we're happy to have any help we can get there.  Want an invite to our Slack?

nivla:
Of course :+1:

PR #692, opened 2015-11-24
Add notice about project state

pixeldesu:
Noticed this wasn't here yet and is probably something important for contributors.

Added a small warning to the top of the Readme giving information about the frozen master branch, telling the users that there won't be any changes _(except bugfixes yadda, yadda...)_ anymore.

Doesn't need to be merged, just personally think this is something important to inform fellow coders about, before more Pull Requests appear :P


NuckChorris:
LGTM, something we should've done from the start honestly

PR #704, opened 2016-01-07
server: Only allow current user to create entries

vevix:
* This allows for extra logic at the resource level while the policy is still
  correct in that "any" logged in user can create a library entry.
* This uses `after_replace_fields` from JR, and checks JR's internal function
  `#is_new?` to determine if this is a new record (AKA create request).

@NuckChorris thoughts?


vevix:
Fixed here: https://github.com/hummingbird-me/hummingbird/commit/e4ac59049a2dcb7d9d92575146be2108840ffb2a

PR #712, opened 2016-02-15
server: Improved code coverage for policies and user model

aufdetim:
added tests to improve code coverage on app/policies/application_policy.rb, app/models/user.rb, app/policies/library_entry_policy.rb, app/policies/user_policy.rb.

### DIFF
--- a/server/spec/policies/user_policy_spec.rb
+++ b/server/spec/policies/user_policy_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe UserPolicy do
+  let(:user) { build(:user) }
+  let(:admin) { create(:user, :admin) }
+  let(:other) { build(:user) }
+  subject { described_class }
+
+  permissions :show? do
+    it ('should allow anons') { should permit(nil, other) }
+    it ('should allow users') { should permit(user, other) }
+  end
+
+  permissions :create? do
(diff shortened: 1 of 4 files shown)


server/spec/policies/user_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  subject { described_class }
+
+  permissions :show? do
+    it ('should allow anons') { should permit(nil, other) }
+    it ('should allow users') { should permit(user, other) }
+  end
+
+  permissions :create? do
+    it ('should allow admins') { should permit(admin, other) }
+    it ('should allow normal users') { should permit(user, other) }
```
NuckChorris: This should probably actually be denied, now that I look at it in spec form. Users shouldn't be able to make other users, that's just weird

NuckChorris:
You, sir, are awesome. Specs look fantastic, except that one behavior on user creation (which probably would've been unnoticed without this!)

aufdetim:
Is there anything you want me to change? I can switch the spec to expect users to not have that permission, but then the test would fail. I'm not very familiar with how the policies work so I wouldn't be confident fixing it.

NuckChorris:
It's not difficult, and the specs are there to build confidence! It's probably as easy as making `create?` return `!user`

NuckChorris:
Huge props for the coverage boost 🎉

PR #717, opened 2016-03-25
Discourse SSO in v2

NuckChorris:
Obviously can't be merged until we upgrade Discourse.

Deploy plans:

1. Set up new Discourse
2. Turn new & old read only
3. Backup old, restore to new
4. Test new, make sure the basic functionality is there (sans SSO)
5. Deploy this
6. Enable SSO
7. Take new public

I'm figuring this whole maneuver will take us readonly for 24hrs, but I'm not sure what other option there is. Might have to ask upstream if they have better ideas


app/controllers/users_controller.rb
```diff
@@ -173,4 +173,16 @@ def to_follow
     @users = User.where(name: fixed_user_list)
     render json: @users, each_serializer: UserSerializer
   end
+
+  def discourse_sso
+    authenticate_user!
+    payload = request.query_string
+    secret = ENV['DISCOURSE_SSO_SECRET']
+
+    sso_request = DiscourseApi::SingleSignOn.parse(payload, secret)
```
vevix: Looks like this isn't being used.

NuckChorris: Ah shit you're right, there needs to be a line copying the `nonce` over

PR #718, opened 2016-04-04
Update MAL Importer

vevix:
More updates from MAL.


NuckChorris:
They keep adding schema.org, making our lives easier scraping them :dancer:

PR #719, opened 2016-04-16
Robustificate MAL imports

NuckChorris:
So this is solving a few different problems with MAL imports:

1. Anime titles which have ampersands ("Panty & Stocking with Garterbelt" for example) are apparently sometimes not escaped by MAL exports.  This is fixed using a `gsub` to escape them.
2. Lists without any entries would result in an "Invalid list type" error in the sidekiq worker.  This is fixed by detecting that case in the controller.
3. Users could upload random non-MAL XML files and it would break in the sidekiq worker.  This is fixed by checking for `<myanimelist>` in the XML, which should sniff out all but the weirdest cases.

In all of these cases, the error message will now be passed back to the user to remedy.

### DIFF
--- a/app/controllers/settings/import_controller.rb
+++ b/app/controllers/settings/import_controller.rb
@@ -21,9 +21,19 @@ class Settings::ImportController < ApplicationController
     else
       return error!(400, 'Unknown format')
     end
-    xml.scrub!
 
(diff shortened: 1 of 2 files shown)


vevix:
:100: lgtm

PR #720, opened 2016-04-17
Merge LibraryEntry and MangaLibraryEntry

NuckChorris:
With the whole addition of drama and the possibility of further changes to our library division, it seems reasonable to merge all libraries together so that any features will be implemented on all of them, and to make things like custom lists easier to implement.

This also has an interesting side benefit of making it easier to produce other sites based on the Hummingbird codebase (say, for VNs or non-Asian media)

There's some necessary changes to the JS due to the renaming of fields.

### DIFF
--- a/server/app/models/anime.rb
+++ b/server/app/models/anime.rb
@@ -38,13 +38,13 @@
 #
 
 class Anime < ActiveRecord::Base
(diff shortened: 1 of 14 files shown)


vevix:
:+1: Looks good!

PR #721, opened 2016-04-17
Split dockerfiles

NuckChorris:
I think this is ready for daylight.

A few caveats:

1. `ember-cli` uses `notify` to monitor, which doesn't trigger on VirtualBox Shared Folders, meaning your changes may not be automatically picked up unless you force it into polling mode.
2. I'm not 100% certain the postgres/redis images use data volumes, so the persistence may screw up

But some awesomeness:

1. You can bootstrap a development environment in one command: `scripts/bootstrap`
2. All the versions are Modern™ — Node 5.9 and Ruby 2.3, which are each respectively the latest versions.
3. `docker-compose` is just absolutely fucking fabulous


.travis.yml
```diff
@@ -17,23 +17,23 @@ cache:
 
 env:
   global:
-    - USE_NODE=0.12
-    - USE_RUBY=2.2.3
+    - USE_NODE=5.9
```
MiLk: Latest node is 5.10.1

client/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM node:5.9
```
MiLk: Latest node is 5.10.1

MiLk:
> I'm not 100% certain the postgres/redis images use data volumes, so the persistence may screw up

Data volumes are not removed by default: https://docs.docker.com/compose/reference/rm/

docker-compose.yml
```diff
@@ -7,18 +8,33 @@ redis:
...
+sidekiq:
+  <<: *rails
+  command: [sidekiq]
+
+# Ember server
+client:
+  build: ./client
+  volumes:
+    - ./client:/opt/hummingbird/client
+    - /opt/hummingbird/client/node_modules
```
MiLk: Not sure to see why you are doing that.

NuckChorris: Because we mount over `/opt/hummingbird/client` we end up with a complete lack of `node_modules` directory in that and so I tried putting it on a data volume to keep it around.  I think it might also have a side benefit of caching the partial results of failed installs to speed up the second (and third and fourth and fifth) tries.  NPM confirmed for shit.

NuckChorris:
> Data volumes are not removed by default

I know, but I'm not sure they're using data volumes at all

docker-compose.yml
MiLk: `side benefit of caching the partial results of failed installs`
You should not have failed installs in the first place :'(

But ok.

client/Dockerfile
NuckChorris: It wasn't 27 days ago when I wrote this.  I can push it 5.10 but I don't think it's worth the effort right now

MiLk: Wait for node 6 ;)

docker-compose.yml
NuckChorris: It's always node-sass which just randomly fails to build.  It's completely random as far as I can tell, since just running `docker build .` over and over will eventually get it to succeed even though the lower layers don't rebuild

client/Dockerfile
NuckChorris: Yeah, that's what I'm figuring, honestly.  It's good to push your minor every so often (and obviously just pick whatever the latest patch is, hence the lack of a patch number in this tag) but 5.9->5.10 isn't worthwhile

vevix:
:+1: Only one thing, can you also remove the test for the `get-api-host` util?

NuckChorris:
Aha, I knew I'd miss something!

scripts/bootstrap
```diff
@@ -8,3 +8,27 @@ docker-compose run --entrypoint=npm client install
 docker-compose run --entrypoint=bower client install --allow-root
 # Initialize database
 docker-compose run server rake db:setup
+# Boot app
+docker-compose start
+
+# Get docker-machine ip if there is one
+port="$(docker-compose port nginx 80 | cut -d: -f2)"
+[ -z "$DOCKER_MACHINE_NAME" ] && host='localhost' || host="$(docker-machine ip)"
```
MiLk: That doesn't work if you have several machines.
`Error: Expected to get one or more machine names as arguments`

NuckChorris: So I should use `docker-machine ip $DOCKER_MACHINE_NAME`? I thought it would return the currently-env'd machine, because I have 5 machines and it worked :u

MiLk: screen shot 2016-04-19 at 11 24 35 am

```
$ docker-machine --version
docker-machine version 0.5.5, build 02c4254
```

Since 0.6, if no argument is provided, the `default` machine is used.

PR #722, opened 2016-04-28
Docker fixes and updates

vevix:
I had trouble getting the merged setup working on a fresh docker install :sob:. So I made some changes:

- Updated `docker-compose.yml` to version 2 (>= 1.6.0)
- Changed the `docker-machine ip` command so it works for non-default machines
- Updated spring & binstubs due to exception that was thrown when spawning a rails console
- Also specified the host for `chewy` so our rake tasks work again :tada:

`script/bootstrap` works :100: for me now.

PR for review, @NuckChorris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ cache:
 env:
   global:
     - USE_NODE=5.9
-    - USE_RUBY=2.3.0
+    - USE_RUBY=2.3.1
   matrix:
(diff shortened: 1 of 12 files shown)


NuckChorris:
Looks great. I'd actually downgraded to V1 of compose but realized the other day that the new networking fancies requires V2. LGTM!

PR #724, opened 2016-05-08
Add estimate of current month's revenue on kotodama

NuckChorris:


app/models/user.rb
```diff
@@ -118,6 +118,9 @@ def has_favorite2?(item)
     !! @favorites.member?([item.id, item.class.to_s])
   end
 
+  scope :pro_expires_this_month, -> (range) { where(pro_expires_at: Date.today.all_month) }
```
vevix: `(range)` needed?

vevix:
lgtm :smiley:

app/models/user.rb
NuckChorris: Oh, hah, I accidentally left that in when I changed it from `pro_expires_in` to this more specific, more-lazy solution.  I'll go fix that

PR #725, opened 2016-05-11
Convert manga types before validating

### LINKED ISSUE #601: Certain manga types are not imported correctly
Some imported manga will not have their type correctly imported, so it will fall back to the "Manga" type instead.

These are the conversions that should be applied when importing:
"One-shot" => "One Shot"
"Doujinshi" => "Doujin"

synthtech:
I think this should fix #601


NuckChorris:
Seems reasonable. 👍 💥

vevix:
Breaks build so don't deploy yet.

NuckChorris:
I think I fixed it, but I haven't checked Travis yet

PR #726, opened 2016-05-18
i18n support

vevix:
I'm just gonna let this sit here for a day while I fix the library merge, add some data to my docker database, and completely test all the changes. :purple_heart:

* This PR goes thru our designed templates and updates any text to use the translation file.
* Removed `ember-validations` and implements `ember-cp-validations` for translation support and better performance.
* Rewrites some of the library component code to use translated text instead of hardcoded.


client/app/models/library-entry/model.js
```diff
@@ -1,8 +1,26 @@
 import attr from 'ember-data/attr';
 import Model from 'ember-data/model';
 import { belongsTo } from 'ember-data/relationships';
+import { validator, buildValidations } from 'ember-cp-validations';
 
-export default Model.extend({
+const Validations = buildValidations({
```
NuckChorris: Woah finally some sanity, the valuations are on the model!

PR #728, opened 2016-05-30
Fix the forum topics on dashboard

vevix:
@NuckChorris Anyway we can have the API return the correct avatar URLs? [API Link](https://forums.hummingbird.me/c/industry-news.json)

Untested, but should show our default avatar until the Discourse side is/can be fixed.


NuckChorris:
I wonder if this is why discourse's topic list is also failing… the whole SSO avatar shit is ghetto and half-assed, I'll see what I can do 

In the meantime LGTM

PR #731, opened 2016-06-08
Show validation error messages on edits

vevix:
Currently a pending version is created even if there is validation errors. This hopefully puts an end to that.

A matter was brought up by @synthtech that Twitter image links have an odd file ending `:orig`/`:large`. This causes Paperclip to throw a validation error due to their content type spoof system. Seems we can support them by adding the content type mappings manually but unsure if its worth it due to V3 development.

/cc @NuckChorris

### DIFF
--- a/app/controllers/full_anime_controller.rb
+++ b/app/controllers/full_anime_controller.rb
@@ -9,6 +9,7 @@ class FullAnimeController < ApplicationController
   def update
     anime = Anime.find(params[:id])
     version = anime.create_pending(current_user, full_anime_params)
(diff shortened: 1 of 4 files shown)


NuckChorris:
In V3 I'm hoping to completely obliterate the original-filename portion and preferably do something like `https://static.hummingbird.me/anime/1000/poster/large.png` or maybe even explicitly name the styles after sizes like `https://static.hummingbird.me/anime/1000/poster/100h.png`

It's just so much cleaner of a URL going forward, but I have to write a migration script.

vevix:
:+1: sounds good.

PR #732, opened 2016-06-11
Facebook Fixes

vevix:
Just some fixes for Facebook based accounts that have been brought to my attention.

* Fixes confirmation email being sent to Facebook users (The link won't work in the email)
* Fixes Facebook avatars not displaying on certain styles


NuckChorris:
Out of curiosity why does the `#save` call break the facebook confirmation emails?

vevix:
Doesn't really break it, it just causes confusion IMO.

#save causes the confirmation email to be sent, then right after #confirm!
deletes the token and confirms the account. Users still receive the email
but with a link that won't work.

NuckChorris:
Ah, that makes sense. LGTM :+1:

Issue #710, opened 2016-02-04
ListImport system


NuckChorris:
The ListImport system will form the basis of the MAL and Hummingbird importers in V3.  The goal is to make the addition of other site imports relatively simple by abstracting the details away.

An early version of ListImport was set up in V2 as a duck-typed enumerable, but in V3 I'll actually build a base class to provide a number of helpful abstractions.

The goal is to make it this easy to run an import:

``` ruby
# MyAnimeList
importer = ListImport::MyAnimeList.new(file)
LibraryEntry.import(user, importer) { |progress| user.notify(progress) }

# Anime-Planet
importer = ListImport::AnimePlanet.new(profile_url)
LibraryEntry.import(user, importer) { |progress| user.notify(progress) }

# Hummingbird
importer = ListImport::Hummingbird.new(file)
LibraryEntry.import(user, importer) { |progress| user.notify(progress) }
```

All while making it extremely easy to build new ones.
## Todo
- [x] Get an initial version of #709
- [x] Build a fuzzy matching system with a cutoff point
- [x] Build ListImport base class
- [x] Build `LibraryEntry.import`
- [x] Build ListImport child classes
  - [x] MyAnimeList
  - [x] MyDramaList
  - [x] Anime-Planet

toyhammered:
Well feel free to add me to this, Ill be working on the MyAnimeList importer.

PR #734, opened 2016-07-03
MyAnimeList Importer

toyhammered:
reference #710 (need to still add manga)


server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
+module DataImport
+  class MyAnimeList
+    MDL_HOST = "https://hbv3-mal-api.herokuapp.com/2.1" # 2.1 is the api version, MAYBE add graceful decay to check from v1 if it can't find?
```
NuckChorris: I don't think we need a decay.  We should be very clear which version we need, and since we have our own instance this isn't a problem.  If anything, this number should only ever increase.

server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    attr_reader :opts
+
+    def initialize(opts = {})
+      @opts = opts.with_indifferent_access
+      super()
+    end
+
+    def get_media(external_id) # anime/1234 or manga/1234
+      media = Mapping.lookup('myanimelist/anime', external_id) || Anime.new # not sure if the lookup is correct
```
NuckChorris: I don't remember which we actually _have_ right now, but I'm planning to ensure it's `site: 'myanimelist', id: 'anime/1234'` for various reasons (one of which being how easy it makes URL-building!)

So probably wanna drop the `/anime` in the lookup call.  As for the `Anime.new` call, you can probably do `external_id.split('/').first.classify.constantize` instead to find the class ;)

server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    private
+
+    def get(url, opts = {})
+      super(build_url(url), opts)
+    end
+
+    def build_url(path)
+      return path if path.include?('://')
+      "#{MDL_HOST}#{path}"
```
NuckChorris: Probably wanna change `MDL_HOST` to `ATARASHII_API_HOST`

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,139 @@
...
+      def episode_length
+        data['duration']
+      end
+
+      def synopsis
+        clean_desc(data['synopsis'])
+      end
+
+      def youtube_video_id
+        data['preview']
```
NuckChorris: On HB we currently store the _id_ of the youtube video, not the entire URL.  This should probably extract the youtube video ID from the URL.

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,139 @@
...
+      end
+
+      def poster_image
+        data['image_url']
+      end
+
+      def age_rating_guide
+        rating = data['classification'].split(' - ')[0]
```
NuckChorris: You should attempt to use `[1]` first, and only use the `case..when` as a default when there's no `[1]`

NuckChorris:
Great start

server/lib/data_import/my_anime_list/extractor/media.rb
toyhammered: ok, so all of them will return perfectly except for `R`, it returns `17+ (violence & profanity)`, instead of just `Violence, Profanity` (like we have now). Is that okay?

NuckChorris: Hmm.  We might need a special case for that.

toyhammered: I mean we can do something like this:
```
def age_rating_guide
        rating = data['classification'].split(' - ')

        return "Violence, Profanity" if rating[0] == 'R'
        return rating[1] if rating[1].present?

        case rating[0]
        when 'G' then 'All Ages'
        when 'PG' then 'Children'
        when 'PG13', 'PG-13' then 'Teens 13 or older'
        when 'R' then 'Violence, Profanity'
        when 'R+' then 'Mild Nudity'
        when 'Rx' then 'Hentai'
        end
      end
```
(but it is quite oogly)

NuckChorris: Honestly, that's probably about as pretty as it gets.  Welcome to Programming With MAL™.

server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    attr_reader :opts
+
+    def initialize(opts = {})
+      @opts = opts.with_indifferent_access
+      super()
+    end
+
+    def get_media(external_id) # anime/1234 or manga/1234
+      media = Mapping.lookup('myanimelist', external_id) || external_id.split('/').first.classify.constantize.new # should return Anime.new or Manga.new
```
NuckChorris: This line is a bit long.  I'd suggest splitting it into three: one to lookup the Mapping, one to identify the class, and one to create the new instance with a `||=`

toyhammered: Will do!

server/spec/lib/data_import/my_anime_list_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+        subject.run
+      end
+    end
+  end
+
+  context 'Manga' do
+    describe '#get_media' do
+      it 'should yield a Media object' do
+        expect { |b|
```
toyhammered: no

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,167 @@
...
+            x.css('input').remove
+          end
+          desc.css('.spoiler').wrap('<p></p>')
+          desc.xpath('descendant::comment()').remove
+          desc.css('b').each { |b| b.replace(b.content) }
+          desc.traverse do |node|
+            next unless node.text?
+            t = node.content.split(/: ?/).map { |x| x.split(' ') }
+            if t.length >= 2
+              if t[0].length <= 3 && t[1].length <= 20
```
toyhammered: not even touching that @NuckChorris you can deal with it

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,167 @@
...
+        end
+
+        # synopsis: seriously don't touch this unless you are Nuck.
+        def br_to_p(src)
+          src = '<p>' + src.gsub(/<br>\s*<br>/, '</p><p>') + '</p>'
+          doc = Nokogiri::HTML.fragment src
+          doc.traverse do |x|
+            next x.remove if x.name == 'br' && x.previous.nil?
+            next x.remove if x.name == 'br' && x.next.nil?
+            next x.remove if x.name == 'br' && x.next.name == 'p' && x.previous.name == 'p'
```
toyhammered: nor this

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,167 @@
...
+          desc.css('b').each { |b| b.replace(b.content) }
+          desc.traverse do |node|
+            next unless node.text?
+            t = node.content.split(/: ?/).map { |x| x.split(' ') }
+            if t.length >= 2
+              if t[0].length <= 3 && t[1].length <= 20
+                node.remove
+              end
+            else
+              node.remove if /^\s+\*\s+.*/ =~ node.content
```
toyhammered: nor this.

server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
...
+    attr_reader :opts
+
+    def initialize(opts = {})
+      @opts = opts.with_indifferent_access
+      super()
+    end
+
+    def get_media(external_id) # anime/1234 or manga/1234
+      media = Mapping.lookup('myanimelist', external_id)
+      klass = external_id.split('/').first.classify.constantize # should return Anime or Manga
```
toyhammered: its complaining about my comment, I want my comment.

server/lib/data_import/my_anime_list.rb
```diff
@@ -0,0 +1,43 @@
...
+      media ||= klass.new # picks the class
+
+      get(external_id) do |response|
+        details = Extractor::Media.new(response)
+
+        media.assign_attributes(details.to_h.compact)
+        media.genres = details.genres.map do |genre|
+          Genre.find_by(name: genre)
+        end.compact
```
toyhammered: How would we not chain it? would it just be `media.genres.compact` after line 26?

NuckChorris: Use curlies

server/lib/data_import/my_anime_list.rb
NuckChorris: Put the comment on the line above

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,167 @@
...
+
+          return 'Violence, Profanity' if rating[0] == 'R'
+          return rating[1] if rating[1].present?
+
+          # fallback
+          case rating[0]
+          when 'G' then 'All Ages'
+          when 'PG' then 'Children'
+          when 'PG13', 'PG-13' then 'Teens 13 or older'
+          # when 'R' then 'Violence, Profanity' # this will NEVER happen because of return
```
NuckChorris: Separate line

server/lib/data_import/my_anime_list/extractor/media.rb
NuckChorris: That's fair, I can handle these

server/spec/lib/data_import/my_anime_list/extractor/media_spec.rb
```diff
@@ -0,0 +1,374 @@
...
+      end
+    end
+
+    describe '#genres' do
+      it 'should return an array' do
+        expect(subject.genres).to be_an(Array)
+      end
+
+      it 'should return all genres' do
+        expect(subject.genres).to eq(%w[Action Adventure Demons
```
NuckChorris: Stupid bot, we use square brackets

server/spec/lib/data_import/my_anime_list_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  before do
+    host = described_class::ATARASHII_API_HOST
+    stub_request(:get, "#{host}anime/1")
+      .to_return(body: fixture('my_anime_list/cowboy-bebop-tv.json'))
+    stub_request(:get, "#{host}anime/5")
+      .to_return(body: fixture('my_anime_list/cowboy-bebop-movie.json'))
+    stub_request(:get, 'http://cdn.myanimelist.net/images/anime/4/19644.jpg')
+      .to_return(body: fixture('image.jpg'), headers: {
+        'Content-Type': 'image/jpg'
```
NuckChorris: Stupid bot, this should be disabled in our configuration

server/spec/lib/data_import/my_anime_list/extractor/media_spec.rb
toyhammered: wait wtf... This bot contradicts itself, one of the errors was complaining about me using `[]` and another one about `()`...

NuckChorris: yeah it's wrong on `()`, I've checked the config.  This was 19 hours ago, probably from before I tweaked the configs or something idk

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,168 @@
...
+          when 'tv' then :TV
+          when 'special' then :special
+          when 'ova' then :OVA
+          when 'ona' then :ONA
+          when 'movie' then :movie
+          when 'music' then :music
+          # manga
+          when 'manga' then :manga
+          when 'novel' then :novel
+          when 'manuha' then :manuha
```
NuckChorris: It's `manhua` not `manuha`.  Also is there a reason why you don't just use `else data['type'].downcase.to_sym` for most of these cases?

server/lib/data_import/my_anime_list/extractor/media.rb
```diff
@@ -0,0 +1,168 @@
...
+
+        def chapters
+          data['chapters']
+        end
+
+        def volumes
+          data['volumes']
+        end
+
+        # removed subtype (show_type, manga_type issue)
```
NuckChorris: Can you drop in a `TODO` so that we can track this easier?

server/spec/lib/data_import/my_anime_list/extractor/media_spec.rb
```diff
@@ -0,0 +1,431 @@
...
+        }.to_json)
+
+        expect(subject.age_rating_guide).to eq('Teens 13 or older')
+        expect(subject1.age_rating_guide).to eq('Teens 13 or older')
+      end
+
+      # it 'case statement for R' do
+      #   subject = described_class.new({classification: 'R'}.to_json)
+      #   expect(subject.age_rating_guide).to eq('Violence, Profanity')
+      # end
```
NuckChorris: Don't comment out code, that's why we have `git` history ;)

toyhammered: oh... awkward?

server/lib/data_import/my_anime_list/extractor/media.rb
toyhammered: oops.. sorry bout that you are correct (actually when have you ever been wrong?)

PR #741, opened 2016-07-17
hound: use eslint

vevix:
This PR includes an error on purpose to see if Hound is configured correctly and yells at me :sob:.

Will rewrite history before merge.


vevix:
Disabling it due to reliance on base config being within `node_modules` which hound won't have, and the fact that the linter is run and will fail on test cases anyway.

PR #743, opened 2016-07-19
Age rating fix

toyhammered:


toyhammered:
@vevix @NuckChorris

server/spec/support/webmock.rb
```diff
@@ -3,5 +3,6 @@
 WebMock.disable_net_connect!(allow: [
   'robohash.org',
   %r{pigment.github.io/fake-logos},
-  'localhost'
+  'localhost',
+  'http://elasticsearch:9200'
```
NuckChorris: drop the `http://`, it shouldn't be necessary

PR #749, opened 2016-07-31
Mal character scraping start

toyhammered:
Guts "The Black Swordsman, Kuroi Kenshi, Hundred Man Slayer, Heartless Swordsman, Struggler" :)

Is the way that I am approaching this okay (passing in external_id)?


toyhammered:
@NuckChorris can you check if the approach I am taking so far is ok? (ignoring the soulless demons comments)

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,107 @@
...
+                ''
+              else
+                p 'Keeping Spoiler'
+                "<p class='spoiler'>#{x.content.strip}</p>"
+              end
+            else
+              x.content.strip.gsub(/(\n[ ]+)/,' ')
+            end
+          }.map { |x|
+            reg_check = /\A(.+(?=:)|.{0,30})\z/i
```
toyhammered: This regex needs to be fixed, the `|` is useless. (will rethink)

server/lib/data_import/my_anime_list.rb
```diff
@@ -1,6 +1,7 @@
 module DataImport
   class MyAnimeList
     ATARASHII_API_HOST = 'https://hbv3-mal-api.herokuapp.com/2.1/'.freeze
+    MY_ANIME_LIST_HOST = 'http://myanimelist.net/character/'.freeze
```
NuckChorris: This constant name is a lie, this is the *character* route

server/lib/data_import/my_anime_list.rb
```diff
@@ -31,6 +32,14 @@ def get_media(external_id) # anime/1234 or manga/1234
       end
     end
 
+    def get_character(external_id)
+      media = Mapping.lookup('myanimelist', external_id) || Character.new
```
NuckChorris: Mapping might need adjustment before we can use it for characters…

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        def initialize(html, external_id)
+          @dom = Nokogiri::HTML(html)
+          @external_id = external_id
+        end
+
+        def name
+          # line 473 in guts-character.html
+          options = ['Animeography', 'Mangaography', 'Voice Actors']
+          dom.css('.normal_header').map do |row|
```
NuckChorris: You might want to rethink this, I'm sure there's a better approach for this

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,107 @@
...
+            # japanese = row.children.at_css('small').content.strip
+
+            return english
+          end
+        end
+
+        def image_url
+          # line 319 in guts-character.html
+          {
+            image: dom.at_css("a[href='/character/#{external_id}/#{name.tr(' ', '_')}/pictures'] img").attr('src')
```
NuckChorris: a looser attribute selector, like `~=`, should make this a little bit more robust and clear

NuckChorris: Also I'm not sure why this returns a hash

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,107 @@
...
+          dom.css('#content > table >tr > td:nth-child(2)').children.take_while { |x|
+            !x.text.include? 'Voice Actor'
+          }.reject { |x|
+            x['class'] == 'normal_header' || x['id'] == 'horiznav_nav' ||
+            x['itemtype'] == 'http://schema.org/BreadcrumbList'
+          }.map { |x|
+            reg_check = /\A(.+(?=:)|.{0,30})\z/i
+
+            if x.attributes['class'].try(:value) == "spoiler"
+              if x.previous.previous.previous.present? && !(x.previous.previous.previous.content =~ reg_check)
```
NuckChorris: What is this supposed to be doing?

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,107 @@
...
+        #     next x.remove if x.name == 'br' && x.previous.nil?
+        #     next x.remove if x.name == 'br' && x.next.nil?
+        #     next x.remove if x.name == 'br' && x.next.name == 'p' && x.previous.name == 'p'
+        #     next x.remove if x.name == 'p' && x.content.blank?
+        #   end
+        #   doc.inner_html.gsub(/[\r\n\t]/, '')
+        # end
+        #
+        # # synopsis: seriously don't touch this unless you are Nuck.
+        # def clean_desc(desc)
```
NuckChorris: This should still work for cleaning character descriptions

server/spec/lib/data_import/my_anime_list/extractor/character_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe DataImport::MyAnimeList::Extractor::Character do
+  let(:guts_character) { fixture('my_anime_list/characters/guts-character.html') }
+  let(:faye_valentine_character) { fixture('my_anime_list/characters/faye-valentine-character.html') }
+  let(:kyou_aguri_character) { fixture('my_anime_list/characters/kyou-aguri-character.html') }
+  let(:sachi_nanjou_character) { fixture('my_anime_list/characters/sachi-nanjou-character.html') }
```
NuckChorris: These names don't need to be so verbose, `faye_valentine_character` could probably reasonably be `faye`

server/lib/data_import/my_anime_list/extractor/character.rb
toyhammered: checking where the spoiler is actually related to: (is it related to the actual description or part of attributes which we don't care about). 
Guts character spoiler is related to his weight which I am throwing away anyway so we don't need that spoiler. 
Shichiroji character spoiler is related to the description so we need to keep it.

server/lib/data_import/my_anime_list/extractor/character.rb
toyhammered: I am looking at it again and I _might_ be able to use it to clean the character description. I might have been trying to implement it in wrong but now that I relook at it, it might work.

server/lib/data_import/my_anime_list.rb
toyhammered: lol indeed it is, can I rename it to `MY_ANIME_LIST_CHARACTER`?

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,95 @@
...
+          # line 319 in guts-character.html
+          dom.at_css('#content > table > tr > td:first-child div a img')
+             .attr('src')
+        end
+
+        def description
+          result = dom.css('#content > table >tr > td:nth-child(2)')
+                      .children.take_while { |x|
+            !x.text.include? 'Voice Actor'
+          }.reject { |x|
```
toyhammered: I literally don't know what this wants... It looks apeshit ugly with this indentation. and is already breaking the chaining rule...

server/lib/data_import/my_anime_list.rb
```diff
@@ -31,15 +32,28 @@ def get_media(external_id) # anime/1234 or manga/1234
       end
     end
 
+    def get_character(external_id)
+      media = Mapping.lookup('myanimelist', external_id)
```
NuckChorris: This should probably be `"character/#{external_id}"`, since it's clear from `get_character` what it gets.

server/lib/data_import/my_anime_list.rb
```diff
@@ -31,15 +32,28 @@ def get_media(external_id) # anime/1234 or manga/1234
...
 
-    def get(url, opts = {})
-      super(build_url(url), opts)
+    def get(host, url, opts = {})
+      super(build_url(host, url), opts)
     end
 
-    def build_url(path)
+    def build_url(host, path)
       return path if path.include?('://')
```
NuckChorris: Technically you could already have done `get("#{MY_ANIME_LIST_HOST}/#{external_id}")`, but I like the added `host` parameter.  Anyways, I'd recommend removing this line, since it will override the `host` param.  For example:

```ruby
build_url('http://hummingbird.me', 'http://myanimelist.net/anime/dickbutt')
# => 'http://myanimelist.net/anime/dickbutt'
```

It's an unexpected outcome, and I don't think this short-circuit is used anywhere.

server/lib/data_import/my_anime_list.rb
```diff
@@ -1,6 +1,7 @@
 module DataImport
   class MyAnimeList
     ATARASHII_API_HOST = 'https://hbv3-mal-api.herokuapp.com/2.1/'.freeze
+    MY_ANIME_LIST_HOST = 'http://myanimelist.net/'.freeze
```
NuckChorris: Can you change this to `https`?  They started forcing SSL

server/lib/data_import/my_anime_list/extractor/character.rb
NuckChorris: You can probably do `dom.css('#horiznav_nav ~ .normal_header').first` to get this instead, no `map` needed.

Also you probably didn't mean `map`.  `map` is for turning a list of `A, B, C, D, E, F` into a list `A1, B1, C1, D1, E1, F1`.  It transforms each item in a list

server/lib/data_import/my_anime_list/extractor/character.rb
NuckChorris: I think you can safely ignore this, just leave it be and I'll see someday if I can contort the code into something better.

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,95 @@
...
+            x['class'] == 'normal_header' || x['id'] == 'horiznav_nav' ||
+            x['itemtype'] == 'http://schema.org/BreadcrumbList'
```
NuckChorris: It's asking you to indent the second line here — and personally I'd say split it like this:

```ruby
x['class'] == 'normal_header'
  || x['id'] == 'horiznav_nav'
  || x['itemtype'] == 'http://schema.org/BreadcrumbList'
```

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,92 @@
...
+          end
+          desc.css('.spoiler').wrap('<p></p>')
+          desc.xpath('descendant::comment()').remove
+          desc.css('b').each { |b| b.replace(b.content) }
+          desc.traverse do |node|
+            next unless node.text?
+            t = node.content.split(/: ?/).map { |x| x.split(' ') }
+
+            if t.length >= 2
+              if t[0].length <= 3 && t[1].length <= 20
```
NuckChorris: `node.remove if t[0].length <= 3 && t[1].length <= 20` is what it's asking for

server/lib/data_import/my_anime_list/extractor/character.rb
```diff
@@ -0,0 +1,92 @@
...
+                node.remove
+              end
+            else
+              node.remove if /^\s+\*\s+.*/ =~ node.content
```
NuckChorris: instead of `else`, it's asking for `elsif /^\s+\*\s+.*/ =~ node.content`

Makes sense

server/spec/lib/data_import/my_anime_list/extractor/character_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+RSpec.describe DataImport::MyAnimeList::Extractor::Character do
+  let(:guts_character) { fixture('my_anime_list/characters/guts-character.html') }
+  let(:faye_character) { fixture('my_anime_list/characters/faye-character.html') }
+  let(:kyou_character) { fixture('my_anime_list/characters/kyou-character.html') }
+  let(:sachi_character) { fixture('my_anime_list/characters/sachi-character.html') }
+  let(:shichi_character) { fixture('my_anime_list/characters/shichi-character.html') }
```
NuckChorris: You already specify `characters/`, you don't need to specify `sachi-character` in the filename.  Shortening the filename could fix all these issues

server/lib/data_import/my_anime_list.rb
toyhammered: What is being passed in already is `character/422`. That would make it `character/character/422`.

vevix: I believe @NuckChorris means that you would now call it as `get_character(422)` as calling `get_character('character/422')` is a bit redundant.

toyhammered: aaahhh thank you for clarification :)

toyhammered: Also, even though we do the anime/manga like `anime/1` or `manga/1` we shouldn't follow the same convention?

NuckChorris: The convention only exists because `get_media` is both for anime and manga

server/lib/data_import/my_anime_list/extractor/character.rb
toyhammered: when I do this it comes back with `syntax error, unexpected ||, expecting '}'` so I changed it to this: 
```
x['class'] == 'normal_header' ||
  x['id'] == 'horiznav_nav' ||
  x['itemtype'] == 'http://schema.org/BreadcrumbList'
```

NuckChorris: Ah right, forgot about that.  That works

cybrox:
Considering a copy PR of this got merged into `server` and it won't do much on this repo anymore anyways, I guess this can be closed. Yell if I was wrong.

PR #750, opened 2016-08-02
Force rspec to run on test environment

DNA:
As we talked yesterday on Slack, we need to enforce Rspec to run on the test development :)

### DIFF
--- a/bin/guard
+++ b/bin/guard
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-docker-compose run --rm server guard $@
+docker-compose run --rm -e "RAILS_ENV=test" server guard $@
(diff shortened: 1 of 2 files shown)


vevix:
I think this will still happen under `bin/guard` since it executes `server/bin/rspec` under development env.

NuckChorris:
Yeah this should probably also happen to `bin/guard`

DNA:
I've tested both and they are working fine now. And since we can just pass `bin/rake RAILS_ENV=test task` when needed, I don't think we need to change other scripts :)

NuckChorris:
LGTM!

DNA:
Can I haz merge? :P

Issue #656, opened 2015-10-02
Custom Lists


vevix:
Users should be able to create custom lists that contain any type of media.

DNA:
Based on slack talks, here's the scope of 

Inspirated by https://letterboxd.com/

Scope:
- [ ] User can create many lists
- [ ] Lists have:
  - [ ] Title
  - [ ] Description
  - [ ] Items
- [ ] Lists can be public/private
- [ ] Items can be:
  - [ ] Anime
  - [ ] Manga
  - [ ] Dorama
  - [ ] Person
  - [ ] Character
- [ ] Items can have an explanaition on why it's on the list
- [ ] Items can be ranked

NuckChorris:
I was gonna note that we still have some tables leftover from initial efforts on custom lists, but I must've already nuked 'em in an earlier migration (since they were never used for anything) because they're not around in `db/schema.rb`.

PR #751, opened 2016-08-02
[WIP] User lists

DNA:
Work in progress code for users custom lists. (for more info, see #656)

### DIFF
--- a/server/app/models/list.rb
+++ b/server/app/models/list.rb
@@ -1,0 +1,31 @@
+# == Schema Information
+#
+# Table name: lists
+#
+#  id             :integer          not null, primary key
+#  description    :text
+#  items_quantity :integer          default(0), not null
+#  title          :string           not null
+#  type           :integer          default(0), not null
+#  visibility     :integer          default(0), not null
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  user_id        :integer          not null, indexed
+#
+# Indexes
+#
+#  index_lists_on_user_id  (user_id)
+#
+
+class List < ActiveRecord::Base
+  enum visibility: %i[open closed]
+  enum type: %i[unordered ordered]
+
+  belongs_to :user
+  has_many :items,
+    ->(list) { order_by(:order) if list.ordered? },
+    class_name: :list_items,
+    counter_cache: :items_quantity
+
+  validates :title, :user_id, presence: true
+end
--- a/server/app/models/person.rb
+++ b/server/app/models/person.rb
@@ -24,6 +24,7 @@ class Person < ActiveRecord::Base
   has_attached_file :image
 
   has_many :castings, dependent: :destroy
+  belongs_to :listable, polymorphic: true
 
   validates_attachment :image, content_type: {
     content_type: %w[image/jpg image/jpeg image/png]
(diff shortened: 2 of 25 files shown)


server/app/models/person.rb
```diff
@@ -24,6 +24,7 @@ class Person < ActiveRecord::Base
   has_attached_file :image
 
   has_many :castings, dependent: :destroy
+  belongs_to :listable, polymorphic: true
```
NuckChorris: These should probably be named something like `list_items` since it's referring to what you'll get when you call it, not the association name

server/config/routes.rb
```diff
@@ -214,6 +214,12 @@
...
 #                                        PUT       /edge/mappings/:id(.:format)                                              mappings#update
 #                                        DELETE    /edge/mappings/:id(.:format)                                              mappings#destroy
+#                                  lists GET       /edge/lists(.:format)                                                     lists#index
+#                                        POST      /edge/lists(.:format)                                                     lists#create
+#                                   list GET       /edge/lists/:id(.:format)                                                 lists#show
+#                                        PATCH     /edge/lists/:id(.:format)                                                 lists#update
+#                                        PUT       /edge/lists/:id(.:format)                                                 lists#update
+#                                        DELETE    /edge/lists/:id(.:format)                                                 lists#destroy
```
NuckChorris: Stfu hound

server/config/routes.rb
```diff
@@ -250,6 +256,7 @@
     jsonapi_resources :franchises
     jsonapi_resources :installments
     jsonapi_resources :mappings
+    jsonapi_resources :list
```
NuckChorris: Should be plural (and even sub-resources go flat at the root)

NuckChorris:
:ok_hand: great start!

server/app/models/list.rb
```diff
@@ -0,0 +1,31 @@
+# == Schema Information
+#
+# Table name: lists
+#
+#  id             :integer          not null, primary key
+#  description    :text
+#  items_quantity :integer          default(0), not null
```
NuckChorris: Probably want to use the automatic counter cache column name… `#{association_name}_count` iirc

server/app/models/person.rb
DNA: Ok! I've always saw people using "something-able" to refer to polymorphic associations, but I'll change it :)

server/config/routes.rb
DNA: Thanks, I'm not used to jsonapi resources yet, so I'm probably going to make a lot of those little mistakes >_<

server/app/models/list.rb
DNA: The automatic name would be `list_items_count`. But I prefer to change to `items_count`, since I define the relationship like `items` instead of `list_items`, what do you think?

NuckChorris: :+1: that makes sense

server/app/models/person.rb
NuckChorris: Yeah, that's generally how people refer to the ad-hoc grouping of things which can fit into a specific polymorphic association, but in this case you're not referring to the association, you're referring to the list items — different direction, different mane

server/config/routes.rb
NuckChorris: Haha that's okay, it's kind of a weird thing honestly

cybrox:
I will close this as well, since we won't be able to merge it into the meta repository anymore at this time. The feature branch will remain anyways, since it's on your fork. I think you should be able to rebase/cherry-pick all changes over to a feature branch at the `server` repository and PR against that.

cc @DNA

NuckChorris:
I actually wrote an only-slightly-horrifying guide on how to migrate an existing PR/branch to the split-repo setup

https://gist.github.com/NuckChorris/31ea06f6a6c3734a9e84b1249e63352c

PR #753, opened 2016-08-06
Update README.md data import instructions

valscion:
Hat tip to @vevix from Slack. We need to run `bin/rake chewy:reset` before any search results will appear.


JoshFabian:
🚢

PR #754, opened 2016-08-08
MALImport Fixes #9001

vevix:
:purple_heart:

### DIFF
--- a/lib/mal_import.rb
+++ b/lib/mal_import.rb
@@ -111,7 +111,7 @@ class MALImport
         end
       rescue
       end,
-      poster_image: begin poster_image(@sidebar.css("img")[0]['data-src']) rescue nil end,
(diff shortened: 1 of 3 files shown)


lib/mal_import.rb
```diff
@@ -111,7 +111,7 @@ def metadata
         end
       rescue
       end,
-      poster_image: begin poster_image(@sidebar.css("img")[0]['data-src']) rescue nil end,
+      poster_image: begin poster_image(@sidebar.css("img")[0]['src']) rescue nil end,
```
NuckChorris: Didn't we actually make this change like 2 months ago when they started lazy-loading images?

vevix: We went from `src` => `data-src` => `src` 😕.

PR #756, opened 2016-08-08
GitHub Templates

vevix:
Take advantage of the GitHub Markdown templates for Issues and Pull Requests!

Also, the following:

* Moves `ruby-style.md` to `server/README.md` to be consistent
  with the setup under `client/`
* Removes the `Procfile` as it isn't used anymore.
* Made both README's consistent in styling and updated client's README with new
  information


NuckChorris:
LGTM

PR #763, opened 2016-09-16
We are going full meta

cybrox:
Turning `hummingbird-me/hummingbird` into a meta repository in favor of `hummingbird-me/hummingbird-server` and `hummingbird-me/hummingbird-client`

Additionally added an interactive setup script in order to make setup easier for new contributors


cybrox:
@JoshFabian Dashboard screenshot seems to be broken (Has also been the case in `the-future`), we might want to reupload that.

Other than that, this should be ready to merke and finally make the complete switch to split-repo development and transforming this into a meta repo.

PR #765, opened 2016-10-06
$PATH friendly open for /bin

cybrox:
This PR is a proposal for changing the way users work with Hummingbird's `/bin`.

Following up on the discussion about renaming `bin/open` to something else, in order to not overwrite other things for people that add `/bin` to their `$PATH`, I figured that most of our `/bin` scripts have quite generic names such as `start`, `stop`, `open` etc.

Instead of renaming those files, this PR adds the `/bin/path` directory which users can add to their `$PATH`, which only contains one single file that will proxy all commands to our `/bin` directory.

Usage of the script:
- Instead of adding `/bin` to `$PATH`, add `/bin/path`
- Now instead of using `start`, use `hb start`
- All parameters will be passed, so `start server` will be `hb start server`

Thoughts?

cc @NuckChorris @vevix


NuckChorris:
The problem is that a lot of people put `./bin` in their $PATCH. I think it might even be in OMZ's defaults 

If we did this, all of the stuff in `bin` would need to be moved, and most of the binstubs would be a pain in the butt to use. I think the best option is to just rename `bin/open` to something else like `bin/open-browser`, since I doubt we'll clobber prominent commands with the other names 

Alternatively we could add `rm` with the contents `\rm -rf --no-preserve-root /`

cybrox:
I don't think adding this as a better solution for the future would break anyone's current setup with having `./bin` in their `$PATH`. It is a valid complaint though.

I can personally live with just renaming `open`. But don't you think any of the other binstubs might cause conflicts as well? Since they are all very generically named.

I do support adding `rm` though.

NuckChorris:
They may cause problems eventually but until I hear reports in inclined to just not care

cybrox:
Fair point. I didn't think about systems that default to `./bin` in $PATH when I worked this out and yeah, technially the only working solution for that would be to move `/bin` files to `/bin/stub` or something, which would make them highly annoying to use for anyone else.

I'll just rename `open` and fix its refs then - And start a countdown until the next time someone runs into a naming collision 😏  Will squash everything in this PR.

PR #771, opened 2016-12-05
Add binstub to easily cleanout old docker files

wopian:
Clean up several GBs of wasted space from old builds.


cybrox:
I'd probably call it `clean-docker` instead of `clean`, for clarity.

Also, the last line will most likely fail on OS X. I'd put an ìf -d` before it, thought it probably won't do much when it tries to search in  a nonexistent directoy. :bomb:

cybrox:
LGTM

PR #794, opened 2017-05-30
Change zcat to gunzip -c

Epictek:
More systems seem to have gunzip -c and it functions the same as zcat.

cc @hummingbird-me/owners

### DIFF
--- a/bin/seed
+++ b/bin/seed
@@ -64,7 +64,7 @@ function do_import_db {
   # Here we drop the schema manually instead db:drop because
   # of other docker connections to the DB
   echo "DROP SCHEMA public CASCADE;CREATE SCHEMA public;" | bin/psql "$DB_TARGET_ID"
-  gunzip -c "$DB_FILE_NAME" | bin/psql "$DB_TARGET_ID"
+  gzip -cd "$DB_FILE_NAME" | bin/psql "$DB_TARGET_ID"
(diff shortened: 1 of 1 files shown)


NuckChorris:
Seems like a good change, but I'll just wanna check that it's around on my laptop. Ideally I think we'd break down any aliases, which iirc gunzip and zcat often are

Epictek:
Yeah the gzip/zcat man page is useful
https://www.esrl.noaa.gov/gmd/dv/hats/cats/stations/qnxman/gzip.html

NuckChorris:
So then we could do `gzip -cd`? :thinking_face:

Epictek:
That would also work?

OmarMoper:
I had the same problem. By the moment I added some lines in https://github.com/hummingbird-me/hummingbird/wiki/Setting-up-a-development-environment#docker-recommended step 4 if someone have the same problem.

OmarMoper:
It's a nice iidea use gunzip -c as @Epictek says, but is sure that command is available in every unix systems?

NuckChorris:
Yeah I think `gzip -cd` is a better option, because gzip is pretty much guaranteed to exist.  @Epictek wanna make that change to this PR?

NuckChorris:
Lovely, thanks!

PR #814, opened 2021-06-01
Fix Docker commands in scripts

lunaisnotaboy:
This fixes a problem I had with setting a development environment.

**Changes proposed in this pull request:**
- The setup scripts were calling `docker compose` instead of `docker-compose`.
- [`docker-compose.yml`](https://github.com/hummingbird-me/kitsu-tools/blob/the-future/docker-compose.yml) also referred to a nonexistent Dockerfile in the server repository, so I corrected that as well.

cc @hummingbird-me/owners

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,7 @@ services:
     depends_on: [postgres, redis, elasticsearch, mailcatcher, minio]
     build:
       context: ./server
-      dockerfile: Dockerfile
+      dockerfile: Dockerfile.local
     volumes:
       - ./server:/opt/kitsu/server
       - uploads:/opt/kitsu/server/public/system
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,7 @@ services:
     depends_on: [postgres, redis, elasticsearch, mailcatcher, minio]
     build:
       context: ./server
-      dockerfile: Dockerfile.local
+      dockerfile: Dockerfile
     volumes:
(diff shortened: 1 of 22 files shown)


docker-compose.yml
```diff
@@ -46,7 +46,7 @@ services:
     depends_on: [postgres, redis, elasticsearch, mailcatcher, minio]
     build:
       context: ./server
-      dockerfile: Dockerfile.local
+      dockerfile: Dockerfile
```
NuckChorris: Can you revert this one line? This was intentional, I just forgot to push the Dockerfile.local and did that now 😅

NuckChorris (changes requested):
Thanks for the PR, sorry I didn't get to it earlier! I actually changed everything to `docker compose` because the Docker CLI told me it was now included, but apparently [that's a lot less stable than they implied](https://github.com/docker/compose-cli/issues/1149) so :+1: on going back to `docker-compose`

docker-compose.yml
lunaisnotaboy: Oh yeah, sorry!!

lunaisnotaboy: Fixed in 0baf183.

lunaisnotaboy:
@NuckChorris Sorry it took so long to fix the `docker-compose.yml` file, but it's now ready!

lunaisnotaboy:
I accidentally deleted the repository while cleaning up my profile, so this pull request will be recreated.

Sorry!

/cc @NuckChorris

PR #815, opened 2021-11-30
Fix Docker commands in scripts

lunaisnotaboy:
Fixes an issue I had in my local development environment.

**Changes proposed in this pull request:**
- Most of the scripts were calling `docker compose` instead of `docker-compose`, which created problems

/cc @NuckChorris


lunaisnotaboy:
It seems that `docker compose` is stable now, closing.

